PR #26, opened 2017-09-12
Update endpoints

### LINKED ISSUE #12: Add Update Pet Endpoint
https://github.com/ariesprayoga/swagger-java-sample-app/blob/master/src/main/java/com/wordnik/swagger/sample/resource/PetResource.java

The Update Pet Endpoint should return a 404 Not Found if the pet being update does not already exist.

pauljamescleary:
Closes #12

### DIFF
--- a/src/main/scala/io/github/pauljamescleary/petstore/validation/PetValidationInterpreter.scala
+++ b/src/main/scala/io/github/pauljamescleary/petstore/validation/PetValidationInterpreter.scala
@@ -1,26 +1,34 @@
 package io.github.pauljamescleary.petstore.validation
 
+import cats.data.EitherT
 import cats.effect.IO
 import io.github.pauljamescleary.petstore.model.Pet
 import io.github.pauljamescleary.petstore.repository.PetRepositoryAlgebra
 
 class PetValidationInterpreter(implicit repository: PetRepositoryAlgebra[IO]) extends PetValidationAlgebra[IO] {
-  import cats.syntax.monadError._
 
-  def doesNotExist(pet: Pet): IO[Unit] = {
-    repository.findByNameAndCategory(pet.name, pet.category).ensure(PetAlreadyExistsError(pet)) { matches =>
-      matches.forall(possibleMatch => possibleMatch.bio != pet.bio)
-    }.map(_ => ())
+  def doesNotExist(pet: Pet): EitherT[IO, ValidationError, Unit] = EitherT {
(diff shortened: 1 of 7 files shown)


src/main/scala/io/github/pauljamescleary/petstore/validation/PetValidationInterpreter.scala
```diff
@@ -12,4 +12,15 @@ class PetValidationInterpreter(implicit repository: PetRepositoryAlgebra[IO]) ex
       matches.forall(possibleMatch => possibleMatch.bio != pet.bio)
     }.map(_ => ())
   }
+
+  def exists(petId: Option[Long]): IO[Unit] = {
```
daddykotex: curious: any reason you decide to `raiseError` instead of using something like `Either` to bring up the fact that the pet was not found?

pauljamescleary: Your question hints a little at how I structured validations so far.

As of right now, the idea is that validations are `IO[Unit]`, where we assume that we will get a failure inside of an `IO` if the validation fails; otherwise we get a successful result which is always `Unit`

I am most definitely open to other ways to do business validations.

I decided not to use `Validated` at this point because I am not interested in aggregating errors ala applicatives.

So, to answer your question directly, I am using `IO.raiseError` to "lift" the error into the `IO` monad

pauljamescleary: In some further conversations, I believe that the validation effect type is incorrect.  We should be using an `EitherT[IO, BusinessError, A]` as the effect type for validations.

We need to do a natural transformation in the service for lifting

I have some works to do on this part.  Welcome to ideas as I do some codeplay.

daddykotex: Sounds good to me, I must say I'm experimenting with all of this as well.
I was mostly referring to something like you suggest in your second comment. I'll let you know if I think of something worth sharing.

Thanks

pauljamescleary:
@daddykotex curious as to your feedback here

pauljamescleary:
The feedback I got is to use MonadError.  I am still working on an update that uses that.

daddykotex:
That's exactly what I had in mind, good work here

PR #27, opened 2017-09-18
Removed things that need not be implicits

pauljamescleary:
There was no need to have implicits for the class variables in this example.  Removing them as they don't add any value here.

### DIFF
--- a/src/main/scala/io/github/pauljamescleary/petstore/Server.scala
+++ b/src/main/scala/io/github/pauljamescleary/petstore/Server.scala
@@ -11,11 +11,11 @@ import org.http4s.util.StreamApp
 
 object Server extends StreamApp[IO] {
 
-  private implicit val petRepo = DoobiePetRepositoryInterpreter()
-  private implicit val petValidation = new PetValidationInterpreter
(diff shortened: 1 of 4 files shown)


src/main/scala/io/github/pauljamescleary/petstore/Server.scala
```diff
@@ -11,11 +11,11 @@ import org.http4s.util.StreamApp
...
-  private implicit val petRepo = DoobiePetRepositoryInterpreter()
-  private implicit val petValidation = new PetValidationInterpreter
-  private implicit val petService = new PetService[IO]
-  private implicit val orderRepo = DoobieOrderRepositoryInterpreter()
-  private implicit val orderService = new OrderService[IO]
+  private val petRepo = DoobiePetRepositoryInterpreter()
+  private val petValidation = new PetValidationInterpreter(petRepo)
+  private val petService = new PetService[IO](petRepo, petValidation)
+  private val orderRepo = DoobieOrderRepositoryInterpreter()
+  private val orderService = new OrderService[IO](orderRepo)
```
daddykotex: 👍

PR #33, opened 2017-10-29
Feedbacks and comments

fanf:
Hello, 

I find you project extremelly interresting and learn a lot from it. I would have to have that kind of example a couple of years ago. 
When reading it, I show a couple of things that you may find interesting. I'm in no case a functionnal programmation expert, and so take all of that with a big grain of salf. 

The PR actually don't compile, which is kind of fabulous for a PR trying to give feedbacks :) 
The problem is due to warn on unused / fatal warn. I'm not sure why scalac doesn't see some uses of doobie mapper now. 

There is mainly 2 things:

- use a type like Etiher[Error, T] for method that can fail (typically parsing strings),
- put sealed trait children in the corresponding companion object,  

And something which has nothing to do with FP: better using date time with timezone (ZZ in format), because this is an infinite source of pain. 

Hope it helps, and thank you very much for that project !

### DIFF
--- a/src/main/scala/io/github/pauljamescleary/petstore/endpoint/JodaDateTime.scala
+++ b/src/main/scala/io/github/pauljamescleary/petstore/endpoint/JodaDateTime.scala
@@ -5,7 +5,8 @@ import org.joda.time.DateTime
 import org.joda.time.format.DateTimeFormat
 
 object JodaDateTime {
-  private val dateFormat = "yyyy-MM-dd'T'HH:mm:ss.SSSSSS"
+  // you should always add ZZ for the timezone :)
+  private val dateFormat = "yyyy-MM-dd'T'HH:mm:ss.SSSSSSZZ"
   implicit val decodeDateTime: Decoder[DateTime] = Decoder.instance { cursor =>
     cursor.as[String].flatMap { dateTime =>
       Right(this.fromString(dateTime))
@@ -17,6 +18,13 @@ object JodaDateTime {
   }
 
   def toString(dateTime: DateTime): String = dateTime.toString(dateFormat)
+
+  // from string can fail. You should never authorize your code to throw exception
+  // (appart from "system" error. What is "system" (i.e, out of the scope errors is depending
+  // from you app level. For ex, generaly, all Thread related exception are most likelly out
+  // of the scope of your business logic and you don't want to manage them. But it is totally ok
+  // for Akka/monix/etc to take care of them and transform them on "business" error. Or when
+  // you are in a IO stack)
   def fromString(dateTime: String): DateTime =
     DateTime.parse(dateTime, DateTimeFormat.forPattern(dateFormat))
 }
--- a/src/main/scala/io/github/pauljamescleary/petstore/model/OrderStatus.scala
+++ b/src/main/scala/io/github/pauljamescleary/petstore/model/OrderStatus.scala
@@ -1,20 +1,66 @@
 package io.github.pauljamescleary.petstore.model
 
 sealed trait OrderStatus extends Product with Serializable
-case object Approved extends OrderStatus
-case object Delivered extends OrderStatus
-case object Placed extends OrderStatus
 
+/*
+ * Putting case object in the object with the same name
+ * as you sealed trait is a common practice that allows
+ * - to qualify named (better discoverability for other dev)
+ *   ex: OrderStatus.Approved is self-telling, don't need to
+ *   wonder if it is not Status.Approved)
+ * - to use libraries which needs that convention
+ *   (ex: sealerate is a neat little tool that I used frequently
+ *   when I need to iterate on all cases of the ADT)
+ * - it is widelly use in scalaz/monix/shapeless/etc to better
+ *   managed implicits scoping
+ */
 object OrderStatus {
-  def apply(name: String): OrderStatus = name match {
+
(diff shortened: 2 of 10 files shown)


src/main/scala/io/github/pauljamescleary/petstore/endpoint/JodaDateTime.scala
```diff
@@ -5,7 +5,8 @@ import org.joda.time.DateTime
 import org.joda.time.format.DateTimeFormat
 
 object JodaDateTime {
-  private val dateFormat = "yyyy-MM-dd'T'HH:mm:ss.SSSSSS"
+  // you should always add ZZ for the timezone :)
+  private val dateFormat = "yyyy-MM-dd'T'HH:mm:ss.SSSSSSZZ"
```
daddykotex: IMO, we should probably use java.time from 1.8, but it's minor.

In my experience, most of the problems with date, time and timezone is because we use them to store an instant in time. time zone depends on the user preference most of the time. What you want in most of the case is something like: the event occurred on 2017-10-31T12:26:26 America/New York and that date time should be translated to a epoch or some format that is timezone agnostic.

Then, when displayed to the user, the user timezone is used to properly format the date time

src/main/scala/io/github/pauljamescleary/petstore/model/OrderStatus.scala
```diff
@@ -1,20 +1,66 @@
...
+   * So you may want to use "final case object ...", that can
+   * give the jvm better insight on optimisation.
+   */
+  case object Delivered extends OrderStatus
+  //given as example: javap OrderedStatus$Placed
+  //  public final class io.github.pauljamescleary.petstore.model.OrderStatus$Placed$
+  final case object Placed extends OrderStatus
+
+  //here, you can call
+  def allValues = ca.mrvisser.sealerate.values[OrderStatus]
```
daddykotex: ahh interesting

src/main/scala/io/github/pauljamescleary/petstore/endpoint/JodaDateTime.scala
fanf: You are right, the problem is what is store in db. We should ALWAYS store it in UTC or with the local of the place where is was generated so that we can latter display the correct date/time to the user (in user local). At least here, the user knows what was the timezone. But I agree, it's minor.

pauljamescleary:
Thanks for the input!  Will review.

It does make sense to use java 8+ time stuff, I have started using them in my work projects.

src/main/scala/io/github/pauljamescleary/petstore/model/OrderStatus.scala
```diff
@@ -1,20 +1,66 @@
...
     case Placed => "Placed"
   }
+
+  /*
+   * This one should return an Either because it can fail.
+   * You can define it as a fonction of nameOf // allValues
+   * so that you are sure to never forget to add new values
+   * if you extend OrderedStatus (as you won't have the
+   * protection of total matching that you have in nameOf)
+   */
```
pauljamescleary: good points

src/main/scala/io/github/pauljamescleary/petstore/endpoint/JodaDateTime.scala
pauljamescleary: Datetimes I always store as UTC in the database, and then it is up to the app to determine how to present them.

I like this update

pauljamescleary: Seems like maybe circe has first class support for Instant, will check it out

pauljamescleary: woot!  Changed JodaDateTime to Instant... :heavy_check_mark:

src/main/scala/io/github/pauljamescleary/petstore/model/OrderStatus.scala
pauljamescleary: I didn't go here just yet.  I tinkered some with Enumeratum as well.  Moving the values inside the companion object didn't play well with circe (at least when I tried it).

I do plan on revisiting the models however, and will address this then

src/main/scala/io/github/pauljamescleary/petstore/model/OrderStatus.scala
```diff
@@ -1,20 +1,66 @@
...
+ * as you sealed trait is a common practice that allows
+ * - to qualify named (better discoverability for other dev)
+ *   ex: OrderStatus.Approved is self-telling, don't need to
+ *   wonder if it is not Status.Approved)
+ * - to use libraries which needs that convention
+ *   (ex: sealerate is a neat little tool that I used frequently
+ *   when I need to iterate on all cases of the ADT)
+ * - it is widelly use in scalaz/monix/shapeless/etc to better
+ *   managed implicits scoping
+ */
```
pauljamescleary: these are really good points.  I am going to revisit the model once I wrap up some bigger pieces like authentication and scalajs.  Specifically, I would like to better enforce invariants, and your comments here are applicable with that scope of work

pauljamescleary:
I've kept this open because there are things I want to get back to when I get a chance fwiw

PR #37, opened 2017-11-23
Added Flyway for database migrations

pauljamescleary:
Flyway migrations are run only for H2 database, otherwise use
the sbt flyway plugin to run the migrations and pass in the JVM args
to point to your database.

Also added PureConfig to load the PetStore configuration.


daddykotex:
👍  LGTM

Thanks, again, this code base helped me

PR #41, opened 2017-12-19
Addressing issue 39. Remove `unsafeRunSync` in Server, and no need for implicit parameter list

### LINKED ISSUE #39: Remove unnecessary `unsafeRunSync`
First of all, apologies for not sending a PR directly.

In any case, there is an `unsafeRunSync` in `Server` to go from `F[Stream[F, A]]` to `Stream[F, A]`: this is unnecessary, prefer `Stream.force(FofStream)` instead.

Even better, avoid having `F[Stream[F, A]]` in the first place, and instead lift the actions returning `F` into Stream using `Stream.eval`, and have the whole `for` be in `Stream`. The reason for this is to allow creation of resources with `Stream.bracket`, which takes a finaliser and guarantees resource safety, and it's a very useful and common pattern when dealing with resources that e.g. need to be closed or cleaned up.

tbrown1979:
Fixes #39 

Please let me know if you're not a fan of the formatting on the for-comp in the `Server.scala` file


src/main/scala/io/github/pauljamescleary/petstore/Server.scala
```diff
@@ -18,26 +18,23 @@ import org.http4s.util.ExitCode
...
+      orderService  =  OrderService[F](orderRepo)
+      exitCode      <- BlazeBuilder[F]
         .bindHttp(8080, "localhost")
         .mountService(PetEndpoints.endpoints[F](petService), "/")
         .mountService(OrderEndpoints.endpoints[F](orderService), "/")
         .serve
-    }
-}
+    } yield exitCode
+}
```
pauljamescleary: Need a newline here

tbrown1979:
@pauljamescleary Thanks for the review. I'm glad to help! I added the newline 😄

PR #42, opened 2017-12-20
Use ApplicativeError instead of fromListUnsafe for NonEmptyList

tbrown1979:
I made it so the query parameter endpoints will handle the case where the `List` of params couldn't be turned into a `NonEmptyList`. In that case we use the `fromEither` to register a failure that we can then handle however we see fit - such as returning a `BadRequest`. 

Please let me know if there's any improvements or changes you'd like to see.

### DIFF
--- a/src/main/scala/io/github/pauljamescleary/petstore/endpoint/PetEndpoints.scala
+++ b/src/main/scala/io/github/pauljamescleary/petstore/endpoint/PetEndpoints.scala
@@ -1,7 +1,7 @@
 package io.github.pauljamescleary.petstore.endpoint
 
+import cats.{ApplicativeError}
 import cats.data.Validated.Valid
-import cats.data._
 import cats.effect.Effect
 import cats.implicits._
 import io.circe._
@@ -16,8 +16,9 @@ import org.http4s.dsl.Http4sDsl
 import org.http4s.{EntityDecoder, HttpService, QueryParamDecoder}
 
 import scala.language.higherKinds
+import scala.util.control.NoStackTrace
 
-class PetEndpoints[F[_]: Effect] extends Http4sDsl[F] {
+class PetEndpoints[F[_]: Effect](implicit AE: ApplicativeError[F, Throwable]) extends Http4sDsl[F] {
 
   /* Necessary for decoding query parameters */
   import QueryParamDecoder._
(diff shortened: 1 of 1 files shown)


src/main/scala/io/github/pauljamescleary/petstore/endpoint/PetEndpoints.scala
```diff
@@ -109,28 +112,28 @@ class PetEndpoints[F[_]: Effect] extends Http4sDsl[F] {
...
         // We have a list of valid statuses, find them and return
-        for {
-          retrieved <- petService.findByStatus(NonEmptyList.fromListUnsafe(statuses))
+        (for {
+          statusesNel <- AE.fromEither(statuses.toNel.toRight(EmptyStatuses))
+          retrieved <- petService.findByStatus(statusesNel)
           resp <- Ok(retrieved.asJson)
-        } yield resp
+        } yield resp).recoverWith {
+        case EmptyStatuses => BadRequest("tag parameter not specified")
```
pauljamescleary: This would be "status parameter not specified"

pauljamescleary: Would be better if it followed the "action"  in other cases...

```
val action = for {
  statusesNel <- AE.fromEither(statuses.toNel.toRight(EmptyStatuses))
  retrieved <- petService.findByStatus(statusesNel)
  resp <- Ok(retrieved.asJson)
} yield resp

action .flatMap { 
 ...
}
```

src/main/scala/io/github/pauljamescleary/petstore/endpoint/PetEndpoints.scala
```diff
@@ -109,28 +112,28 @@ class PetEndpoints[F[_]: Effect] extends Http4sDsl[F] {
...
-      case GET -> Root / "pets" / "findByTags" :? TagMatcher(Valid(Nil)) =>
-        BadRequest("tag parameter not specified")
-
       case GET -> Root / "pets" / "findByTags" :? TagMatcher(Valid(tags)) =>
-        for {
-          retrieved <- petService.findByTag(NonEmptyList.fromListUnsafe(tags))
-          resp <- Ok(retrieved.asJson)
-        } yield resp
+        (for {
+          tagsNel   <- AE.fromEither(tags.toNel.toRight(EmptyTags))
```
pauljamescleary: We do the same thing here twice.  

Honestly, I think this should come out of the box with http4s.  We need a required, mult query param decoder.

Perhaps we can implement our own "RequiredMultiQueryParamDecoderMatcher" and contribute it back to http4s?

tbrown1979: That's a great idea! I'll look into that :)

PR #43, opened 2017-12-21
Add get order

pauljamescleary:
Added get order by id endpoint

Updated paths to use path params instead of query params when addressing entities by id

### DIFF
--- a/src/main/scala/io/github/pauljamescleary/petstore/endpoint/OrderEndpoints.scala
+++ b/src/main/scala/io/github/pauljamescleary/petstore/endpoint/OrderEndpoints.scala
@@ -7,6 +7,7 @@ import io.circe.generic.extras.semiauto._
 import io.circe.syntax._
 import io.github.pauljamescleary.petstore.model.{Order, OrderStatus}
 import io.github.pauljamescleary.petstore.service.OrderService
+import io.github.pauljamescleary.petstore.validation.OrderNotFoundError
 import org.http4s._
 import org.http4s.circe._
 import org.http4s.dsl.Http4sDsl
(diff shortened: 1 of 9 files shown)


src/main/scala/io/github/pauljamescleary/petstore/endpoint/OrderEndpoints.scala
```diff
@@ -39,8 +40,19 @@ class OrderEndpoints[F[_]: Effect] extends Http4sDsl[F] {
...
     }
 
+  private def getOrderEndpoint(orderService: OrderService[F]): HttpService[F] =
+    HttpService[F] {
+      case GET -> Root / "orders" / LongVar(id) =>
+        orderService.get(id).value.flatMap {
+          case Right(found) => Ok(found.asJson)
+          case Left(OrderNotFoundError) => NotFound("The order was not found")
+          case Left(unexpected) =>
+            InternalServerError(s"Unexpected error: $unexpected")
```
pauljamescleary: Note: I don't like this InternalServerError business across all the endpoints.  Need a better http4s way of handling unexpected errors.

pauljamescleary: The reason we need this `Left(unexpected)` is because we need to satisfy all `ValidationError` types or else we will get an error indicating that the match is incomplete.

This is not a problem when there is a single negative outcome from a function call.

However, when we can have multiple negative outcomes from a function call, we get into a larger issue.  

Unclear on how to handle this scenario, but feels like the error type instead of being a trait with all possible values as outcomes, should be a coproduct itself, of the possible negative outcomes?  Perhaps shapeless would be usable in this case?

As an example, imagine we have the following:

```
THIS IS ALL CONTRIVED...

sealed trait ValidationError
case class PetNotFoundError() extends ValidationError
case class PetAlreadyExistsError() extends ValidationError
case class InvalidPetCategoryError() extends ValidationError

def create(pet: Pet): EitherT[F, ValidationError, Pet] = {
  if (exists) Left(PetAlreadyExistsError)
  if (invalidCategory) Left(InvalidPetCategoryError)
  else Right(pet)
}

def createPetEndpoint(service: PetService) = {
  val action = service.create(pet)
  action.value.flatMap { 
    case InvalidPetCategoryError ...
    case PetAlreadyExistsError ...
    case Right(pet) => Ok(pet.asJson)
  }
}
```

The issue is that our function signature `def create(pet: Pet): EitherT[F, ValidationError, Pet] ` indicates that any ValidationError could be returned.  Even though "PetNotFoundError" is not a valid error for the "create" use case.

PR #44, opened 2017-12-23
Cleanup errors

pauljamescleary:
Returning an Either with ValidationError forces the caller
to handle all possible types of ValidationError even if they
are not valid for the specific function call.

Instead of a general ValidationError, return a specific validation
error for the function.

### DIFF
--- a/src/main/scala/io/github/pauljamescleary/petstore/validation/ValidationError.scala
+++ b/src/main/scala/io/github/pauljamescleary/petstore/validation/ValidationError.scala
@@ -4,5 +4,5 @@ import io.github.pauljamescleary.petstore.model.Pet
 
 sealed trait ValidationError extends Product with Serializable
 final case class PetAlreadyExistsError(pet: Pet) extends ValidationError
-final case class PetNotFoundError() extends ValidationError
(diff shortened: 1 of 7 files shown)


src/main/scala/io/github/pauljamescleary/petstore/validation/ValidationError.scala
```diff
@@ -4,5 +4,5 @@ import io.github.pauljamescleary.petstore.model.Pet
 
 sealed trait ValidationError extends Product with Serializable
 final case class PetAlreadyExistsError(pet: Pet) extends ValidationError
-final case object PetNotFoundError extends ValidationError
-final case object OrderNotFoundError extends ValidationError
+final case class PetNotFoundError() extends ValidationError
```
tbrown1979: Why the change from a `case object` over `case class`?

pauljamescleary: The alternative is to define the return as `EitherT[F, PetNotFoundError.type, Pet]`

That isn't too bad, will post an update

Issue #21, opened 2017-09-06
Add Create User Endpoint


pauljamescleary:
https://github.com/ariesprayoga/swagger-java-sample-app/blob/master/src/main/java/com/wordnik/swagger/sample/resource/UserResource.java

pauljamescleary:
Closed by #46

PR #46, opened 2017-12-24
Initial User Support

pauljamescleary:
The plan is to implement Authentication Middleware in order
to authenticate incoming requests across the app.

In order to do that, we need a mechanism to allow users to signup.

Once we have this merged, I can use TSec to implement auth
middleware using JWT, and then authenticate all requests.  That
will be a future PR however.

This PR gives us what we need for the first step.

Note: We should have validation as well.  I will update the service in a future pr


pauljamescleary:
Addresses #21

pauljamescleary:
I added the validations, so I am going to reopen under a different PR

PR #47, opened 2017-12-24
Introduce Users

pauljamescleary:
Added initial user support.

This should allow us to create the Authentication Middleware using TSEC


src/main/scala/io/github/pauljamescleary/petstore/endpoint/UserEndpoints.scala
```diff
@@ -0,0 +1,40 @@
...
+  private def signupEndpoint(userService: UserService[F]): HttpService[F] =
+    HttpService[F] {
+      case req @ POST -> Root / "users" =>
+        val action = for {
+          user <- req.as[User]
+          result <- userService.createUser(user).value
+        } yield result
+
+        action.flatMap {
+          case Right(saved) => Ok(saved.asJson)
```
martin-g: Maybe status 201 (`Created`) would be better ?

PR #54, opened 2017-12-29
Load config and migrate database in a referentially transparent fashion

DStranger:


src/main/scala/io/github/pauljamescleary/petstore/config/PetStoreConfig.scala
```diff
@@ -14,8 +14,10 @@ object PetStoreConfig {
     * return an error.  This should halt the application from starting up.
     */
   def load[F[_]](implicit E: Effect[F]): F[PetStoreConfig] =
-    loadConfig[PetStoreConfig]("petstore") match {
-      case Right(ok) => E.pure(ok)
-      case Left(e) => E.raiseError(new ConfigReaderException[PetStoreConfig](e))
+    E.delay {
+      loadConfig[PetStoreConfig]("petstore") match {
+        case Right(ok) => ok
+        case Left(e) => throw new ConfigReaderException[PetStoreConfig](e)
```
pauljamescleary: I don't know if this is acceptable or not, in general I see "throw" as a red flag and feel like there is a better way to do it.

Is there a different way we can do this without throw?

src/main/scala/io/github/pauljamescleary/petstore/config/DatabaseConfig.scala
```diff
@@ -21,10 +21,12 @@ object DatabaseConfig {
       implicit S: Sync[F]): F[Unit] =
     if (dbConfig.url.contains(":h2:")) {
       xa.configure { ds =>
-        val fw = new Flyway()
-        fw.setDataSource(ds)
-        fw.migrate()
-        S.pure(())
+        S.delay {
```
pauljamescleary: Heh, awesome 👍

src/main/scala/io/github/pauljamescleary/petstore/config/PetStoreConfig.scala
pauljamescleary: This would work (I tested this locally)...


```
    E.delay(loadConfig[PetStoreConfig]("petstore")).flatMap {
      case Right(ok) => E.pure(ok)
      case Left(e) => E.raiseError(new ConfigReaderException[PetStoreConfig](e))
    }
```

DStranger: Yes, I agree, while both cases would work, your variant looks better.

daddykotex:
yeah 👍

pauljamescleary:
:champagne:

PR #57, opened 2018-01-01
WIP Added auth pieces

pauljamescleary:
This is a stateless JWT authenticator using the User Repo
as the identity store.

This is not integrated yet and subject to change.


pauljamescleary:
@zakpatterson I started this, but really got nowhere fast on the authentication piece.  I haven't revisited it as the tsec guys indicated that they had changes in the API.  I was waiting until the dust settled on tsec before jumping back in.

PR #62, opened 2018-01-21
Feature add doobie query typechecking and fix schema

zakpatterson:
Add typechecking
---
The goal of this is to make the doobie queries used checkable against the database schema in test. [Typechecking queries](https://tpolecat.github.io/doobie-cats-0.4.2/06-Checking.html)

The checker just operates on Query0 and Update0 however, so we must break the queries into separate objects (marked private), so that they can be checked. This further allows us to compose them more easily to create our `ConnectionIO`s.

Found schema type mismatch
---
As explained in the commit messages, this exposed a simple type mismatch between the scala implementation and the database schema. The scala implementation expected that database `ID`s would map to scala `Long`, but `SERIAL` in SQL uses SQL Integer, which better maps to Int. So instead we use `BIGSERIAL` in the db to map to scala `Long`.  

Normally this schema migration would be done using a second schema migration, but since the existing migration would be incorrectly matched with the scala implementation, this might confuse people later. 

More
---
This got a little large. I learned a lot just working on this, so I'm happy for any discussion and I'm happy to modify or break this up a little bit. I also broke up some of the queries into reusable fragments. I hadn't done this before but I think it might be a good way to not repeat ourselves when we write `SELECT` clauses that have to list every field we wish to map.

Interested in your thoughts! Thanks for this project.


pauljamescleary:
This is awesome!!  Will dig in tomorrow and review this week (I am on the East Coast US).  Going to first understand what you were doing and then do a little deep dive into doobie as well.

zakpatterson:
Great. Looking forward to your comments. I'm in DC. This has already helped me a lot with the tagless final pattern and scalacheck. So I'm happy to help in any way I can.

pauljamescleary:
Digging in a little this morning since I have some time.

I really like the first 2 commits.  I am a little skittish on the 3rd commit because the combination of things like `select ++ from ++ " NORMAL SQL HERE"` is something that I have to think about.  Having the SQL together in one line makes my SQL mojo feel good.

I wonder if we change the driver from MySQL to PostgresSQL to help with the warnings?

According to the old doobie docs @tpolecat indicated that MySQL had lots of problems with just checks.  He indicated that should use `checkOutput` instead, so perhaps that is another avenue to look at.

But, man, this is awesome @zakpatterson, I am totally hi-jacking this for my own projects.

zakpatterson:
I agree about the overuse of the query fragments. The thing I like most about doobie is that the SQL is right there and we're not sacrificing anything.

`checkOutput` is a new one to me. It only checks the column types to ensure they align into the Query type parameter, without checking the input types for the prepared statement. This seems like a good fallback but we should try to check both wherever possible.  [Covered here](https://tpolecat.github.io/doobie-cats-0.4.2/06-Checking.html#working-around-bad-metadata)

I'm updating the pull request just with the pagination fragment use still there. I like that use still because one can compose pagination with any query rather than repeating queries just for that purpose. Also then the typechecking passes.

zakpatterson:
I'm seeing some errors in the python tests, even though it appears the checks passed. I'll resolve those now.

pauljamescleary:
Looks like this has some conflicts now.  I really like this new one though, looking forward to looking at it sans conflicts

src/main/scala/io/github/pauljamescleary/petstore/infrastructure/repository/doobie/DoobieUserRepositoryInterpreter.scala
```diff
@@ -66,10 +63,7 @@ class DoobieUserRepositoryInterpreter[F[_]: Monad](val xa: Transactor[F])
     }
 
   def deleteByUserName(userName: String): F[Option[User]] =
-    findByUserName(userName).flatMap {
-      case Some(user) => UserQueries.deleteByUserName(userName).run.transact(xa).map(_ => Some(user))
-      case None => none[User].pure[F]
-    }
+    OptionT(findByUserName(userName)).mapFilter(_.id).flatMapF(delete).value
```
pauljamescleary: hey, nice

src/main/scala/io/github/pauljamescleary/petstore/infrastructure/repository/doobie/DoobieOrderRepositoryInterpreter.scala
```diff
@@ -23,30 +21,40 @@ class DoobieOrderRepositoryInterpreter[F[_]: Monad](val xa: Transactor[F])
       dt => java.sql.Timestamp.from(dt)
     )
 
+  def select(orderId: Long): Query0[orders.Order] = sql"""
+    SELECT PET_ID, SHIP_DATE, STATUS, COMPLETE, ID
+    FROM ORDERS
+    WHERE ID = $orderId
+  """.query[orders.Order]
+
+  def update(order : orders.Order) : Update0 = sql"""
```
pauljamescleary: Is it wrong of me to think that "update" is not a "Query"?

I don't have an analog to go from, not sure if there is a better name than OrderQueries, maybe OrderSQL?

src/main/scala/io/github/pauljamescleary/petstore/infrastructure/repository/doobie/DoobieUserRepositoryInterpreter.scala
```diff
@@ -3,70 +3,70 @@ package io.github.pauljamescleary.petstore.infrastructure.repository.doobie
...
     get(userId).flatMap {
-      case Some(user) =>
-        sql"DELETE FROM USERS WHERE ID = $userId"
-          .update
-          .run
-          .transact(xa)
-          .map(_ => Some(user))
-      case None =>
-        none[User].pure[F]
+      case Some(user) => UserQueries.delete(userId).run.transact(xa).map(_ => Some(user))
```
pauljamescleary: This pattern is here a few times now.  Is there a combinator and/or other monad we can use like `OptionT`?

pauljamescleary:
Had a few really minor comments in there.  I really like this PR, what a great contrib, many thanks

src/main/scala/io/github/pauljamescleary/petstore/infrastructure/repository/doobie/DoobieUserRepositoryInterpreter.scala
zakpatterson: Yes, I think this can be simplified in a similar way, I'll send another PR.

zakpatterson:
Think I deleted my fork remote branch by accident. Reopening.

zakpatterson:
H2 is reporting that all fields should be VARCHAR in the metadata for the update commands. This means the update queries simply can't be typechecked the way they are implemented now. I just removed the typechecking for the update queries. I can research more if this is possible in another way.

Also I tried to reorder and squash some of these commits, but I ran into some conflicts I couldn't figure out right away.

src/main/scala/io/github/pauljamescleary/petstore/infrastructure/repository/doobie/DoobieOrderRepositoryInterpreter.scala
zakpatterson: Updated it with XxxSQL naming instead, I agree that seems better.

tpolecat:
The issue seems to be that `REPLACE INTO` doesn't give accurate metadata. If you change it to `UPDATE` the query check works. A solution would be to have distinct insert and update operations ... you usually want to do this anyway because `INSERT` will generate the id.

zakpatterson:
Ah amazing. I had suspected because we are using the MySQL compatibility mode setting with H2, when it ingests that `REPLACE INTO` query which is MySQL specific, H2 probably needs to transform it internally.  That could probably explain why all the input fields to the query pre-transform are expected to be text.

Also just using the "standard" sql things Rob is suggesting would be good for generalization and instruction anyway.

zakpatterson:
Ok, I was able to implement Rob's suggestion of replacing `REPLACE INTO`,with `INSERT` which does work with the typechecking! 

I still need to revise the existing usage of the endpoints, since `update` is still calling `put`.  I can get this some time Thursday.

pauljamescleary:
The issue with changing `REPLACE INTO` with `INSERT` is that for all of our "update" endpoints, we call the `Repository.put`.

This would seemingly fail for updates (unless shame on me if I do not have uniqueness constraints in the database).

We would have to differentiate `insert` vs. `update` in our repositories.

pauljamescleary:
heh, I see @zakpatterson you had the same idea as I did, sorry I missed the end of your comment

zakpatterson:
We can still do manual upserting via update -> exemption -> insert, which will be nice since it will show how to handle these exceptions within a single transaction with doobie.

tpolecat:
Where are you getting the id when you insert a new record? What was the strategy with `REPLACE`-only?

zakpatterson:
The schema has `BIGSERIAL` as its type for the id, so an auto-incrementing biginteger would be generated by the database.

I saw your [fixpoint types talk](https://www.youtube.com/watch?v=7xSfLPD6tiQ), where you discussed not encoding the database id within a the case class representation of the data. Basically if the ids are generated by the database, using `Option[Long]` in the application layer doesn't really capture the various cases of the data (not-inserted, post-inserted, queried, pre-update, etc). That's a really good talk I'm hoping to work on an example of that use case in this project.

Right now the mapping is:

```
CREATE TABLE USERS (
  ID BIGSERIAL PRIMARY KEY,
  USER_NAME VARCHAR NOT NULL UNIQUE
-- ...
);
```
to:
```
case class User(
// ...
    userName: String,
    id: Option[Long] = None
)
```

pauljamescleary:
@tpolecat Unclear, looks like the func test did not check what was happening to the id -` def test_update_pet(pet_store_client):`

Will be sure add a condition to the func check to be certain.

pauljamescleary:
@tpolecat the strategy with `REPLACE INTO` was mainly had no other good validations to showcase that part of it other than stuff that could be taken care of by a SQL database (see if it already exists, see if it doesn't already exist).  So I went with `put` semantics on the repository so I could implement those validations higher up.

zakpatterson:
Without any unique constraints or primary keys in the database schema I think the db would have always been treating REPLACE INTO as INSERT?

Should I continue working on improving the insert/update validation and semantics in this PR? Or would you like to merge this one first as it seems its accomplished its goals under the prior functionality, then I can work on adding more constraints and doing validated insert / update in a clean one?

PR #63, opened 2018-01-21
Forward functional test exit status to travis, fix exit status variable

zakpatterson:
Some simple typos led to the functional tests output result not being forwarded back to travis. So failed functional tests would appear as passed in the travis output on github.


pauljamescleary:
Wow, merging immediately, nice save!

Issue #65, opened 2018-01-28
Fix up REPLACE INTO / put in Repositories


pauljamescleary:
We used `put` and `REPLACE INTO` in our repositories.  Unfortunately, these make it difficult to type check our queries.

We should more formally do `insert` v. `update` as opposed to just `put`

Keep the existing validation checks in the algebra as they demonstrate how to do validations (even though they would theoretically be redundant).

Update the database schema to put unique constraints.

Be sure to handle the situation where the `insert` and `update` fail due to constraints...`insert` would fail if the record already exists, `update` would fail if the record does not exist.

Hopefully this solves the doobie type checking.

pauljamescleary:
@zakpatterson I opened this one as a placeholder as I merged your other PR.

Thanks again for your help!

pauljamescleary:
Closed by #66

PR #66, opened 2018-01-28
Feature crud semantics

zakpatterson:
This adds constraints in the database schema, and replaces `REPLACE INTO` (intended to be PUT/insert+overwrite), to explicit insert or update.

Issue #65


src/main/scala/io/github/pauljamescleary/petstore/domain/pets/PetService.scala
```diff
@@ -17,17 +17,11 @@ class PetService[F[_]](repository: PetRepositoryAlgebra[F], validation: PetValid
   import cats.syntax.all._
 
   def create(pet: Pet)(implicit M: Monad[F]): EitherT[F, PetAlreadyExistsError, Pet] =
-    for {
-      _ <- validation.doesNotExist(pet)
-      saved <- EitherT.liftF(repository.put(pet))
-    } yield saved
+    validation.doesNotExist(pet) *> EitherT.liftF(repository.create(pet))
```
pauljamescleary: Would prefer to use for comprehension syntax and avoid symbolic operators when possible

src/main/scala/io/github/pauljamescleary/petstore/infrastructure/repository/doobie/DoobiePetRepositoryInterpreter.scala
```diff
@@ -68,9 +74,12 @@ class DoobiePetRepositoryInterpreter[F[_]: Monad](val xa: Transactor[F])
     extends PetRepositoryAlgebra[F] {
   import PetSQL._
 
-  def put(pet: Pet): F[Pet] =
+  def create(pet: Pet): F[Pet] =
     insert(pet).withUniqueGeneratedKeys[Long]("ID").map(id => pet.copy(id = id.some)).transact(xa)
 
+  override def update(pet: Pet): F[Option[Pet]] =
+    pet.id.traverse(PetSQL.update(pet, _).run.map(_ => pet)).transact(xa)
+
```
pauljamescleary: This is rather dense.  Is there a newb way of writing this?

src/main/scala/io/github/pauljamescleary/petstore/infrastructure/repository/doobie/DoobieUserRepositoryInterpreter.scala
```diff
@@ -43,9 +48,13 @@ class DoobieUserRepositoryInterpreter[F[_]: Monad](val xa: Transactor[F])
 
   import UserSQL._
 
-  def put(user: User): F[User] =
+  def create(user: User): F[User] =
     insert(user).withUniqueGeneratedKeys[Long]("ID").map(id => user.copy(id = id.some)).transact(xa)
 
+  def update(user: User): F[Option[User]] = OptionT(user.id.flatTraverse(select(_).option)).flatMapF(
+    u => u.id.traverse(UserSQL.update(user, _).run.map(_ => user))
+  ).value.transact(xa)
```
pauljamescleary: same comment here, favoring approachability if possible.

src/main/scala/io/github/pauljamescleary/petstore/domain/users/UserService.scala
```diff
@@ -39,7 +39,7 @@ class UserService[F[_]](userRepo: UserRepositoryAlgebra[F], validation: UserVali
   def update(user: User)(implicit M: Monad[F]): EitherT[F, UserNotFoundError.type, User] =
     for {
       _ <- validation.exists(user.id)
-      saved <- EitherT.liftF(userRepo.put(user))
+      saved <- EitherT.fromOptionF(userRepo.update(user), UserNotFoundError)
```
pauljamescleary: nice!

zakpatterson: yeah that transformation fell out nicely.

zakpatterson:
Thank you for the review. I"m also learning a lot from your review as well as how to present things in a more approachable way.

src/main/scala/io/github/pauljamescleary/petstore/infrastructure/repository/doobie/DoobieUserRepositoryInterpreter.scala
zakpatterson: Agree, this one is dumping way too many concepts into a single expression
.

src/main/scala/io/github/pauljamescleary/petstore/domain/pets/PetService.scala
zakpatterson: I'm trying to apply some judicious use of the other typeclasses besides just `Functor` and `Monad`.

In this case, we really do mean that the output of the left hand side is not passed to the right hand side. That is, the inner type need not be unwrapped from the context in order to process the following expression inside of the same context. `flatMap`, although more familiar to non-cats users, seems imprecise to me because of this. When one uses `flatMap`, one indicates that the inner object must be unwrapped and passed into the following computation.

I would argue in favor of using `*>` for this case, with a comment and a link to what this means, or we could switch to `productR`?  Alternatively we could use `productR` since it's easier to look up, with a comment perhaps that it's aliased as `*>`?  

What do you think?

src/main/scala/io/github/pauljamescleary/petstore/infrastructure/repository/doobie/DoobiePetRepositoryInterpreter.scala
zakpatterson: Is it the usage of `traverse`? Or can we break up the expression somewhat, i.e.
 
```
  override def update(pet: Pet): F[Option[Pet]] = {
    def doUpdate(intoId: Long): ConnectionIO[Pet] = PetSQL.update(pet, intoId).run.map(_ => pet)
    pet.id.traverse(doUpdate).transact(xa)
  }
```

The part I dislike here this is the .`map(_ => pet)` part.

src/main/scala/io/github/pauljamescleary/petstore/domain/pets/PetService.scala
zakpatterson: The other *Service classes have several methods that could be rewritten using `*>`. It reads like a hurdle. If the validation passes over the hurdle, provide the following expression.

pauljamescleary: `productR` might be harder to understand ;)

Thanks for the explanation, that helps me alot.  I really like the hurdle analogy.

Could you add an inline comment when it is used to add context?  

And I would say if you use it in one place, use it in all the places where it makes sense for consistency.

Thanks for the clarification

src/main/scala/io/github/pauljamescleary/petstore/infrastructure/repository/doobie/DoobieUserRepositoryInterpreter.scala
zakpatterson: 2598335 attempts to improve this

src/main/scala/io/github/pauljamescleary/petstore/domain/users/UserService.scala
zakpatterson: I applied the same trick in several places in 7a14f63

pauljamescleary:
I have to wait until tonight, I want to check out this branch and inspect everything much more closely than I can do presently.

zakpatterson:
That's good, I had a couple updates I wanted to add.

src/main/scala/io/github/pauljamescleary/petstore/domain/pets/PetService.scala
zakpatterson: I've looked at the *> usage again, and I reverted it.  I think you're right about the readability. I have another comment in the main thread.

zakpatterson:
So I reverted the usage of *> in the validator --> query flow.  I looked at a few different ways to represent it:

Example:
```
  for {
    _ <- validation.exists(pet.id)
    saved <- EitherT.fromOptionF(repository.update(pet), PetNotFoundError)
  } yield saved
```
This is pretty clear, but it could be simplified as:

```
validation.exists(pet.id) *> EitherT.fromOptionF(repository.update(pet), PetNotFoundError)
```

The semantics of this are a bit more correct because The result of the validation is not used in the following computation.

But what if there are more validations?  For example:

```
for {
    _ <- validation.exists(id)
    _ <- validation.userHasPermissionsToUpdate(id, userId)
    _ <- validation.starsAlignedProperly(id)
    _ <- service.update(user, id)
} yield user
```

This is definitely more readable than 
```
 validation.exists(id) *>
 validation.userHasPermissionsToUpdate(id, userId) *>
 validation.starsAlignedProperly(id) *>
 service.update(user, id)
```

Or perhaps:
```
val vs = validation.exists(id) andThen
 validation.userHasPermissionsToUpdate(id, userId) andThen
 validation.starsAlignedProperly(id))
vs *> service.update(user, id)
```
Even:
```
List(validation.exists(id), 
validation.userHasPermissionsToUpdate(id, userId),
validation.starsAlignedProperly(id)).reduce(_ andThen _) *> 
service.update(user, id)
```

The for comprehension beats all these in my opinion.

src/main/scala/io/github/pauljamescleary/petstore/infrastructure/repository/doobie/DoobiePetRepositoryInterpreter.scala
```diff
@@ -68,9 +74,14 @@ class DoobiePetRepositoryInterpreter[F[_]: Monad](val xa: Transactor[F])
     extends PetRepositoryAlgebra[F] {
   import PetSQL._
 
-  def put(pet: Pet): F[Pet] =
+  def create(pet: Pet): F[Pet] =
     insert(pet).withUniqueGeneratedKeys[Long]("ID").map(id => pet.copy(id = id.some)).transact(xa)
 
+  override def update(pet: Pet): F[Option[Pet]] = {
```
pauljamescleary: remove the override keyword here

src/main/scala/io/github/pauljamescleary/petstore/infrastructure/repository/doobie/DoobieUserRepositoryInterpreter.scala
```diff
@@ -43,9 +48,20 @@ class DoobieUserRepositoryInterpreter[F[_]: Monad](val xa: Transactor[F])
 
   import UserSQL._
 
-  def put(user: User): F[User] =
+  def create(user: User): F[User] =
     insert(user).withUniqueGeneratedKeys[Long]("ID").map(id => user.copy(id = id.some)).transact(xa)
 
+  def update(user: User): F[Option[User]] = {
+    val existingUser : ConnectionIO[Option[User]] = user.id.flatTraverse(select(_).option)
```
pauljamescleary: this is a little confusing.  If the user id is not present, we should just return F(None).

Seems like we are doing a lot of gymnastics in there.  Would something like this suffice?

```
  def update(user: User): F[Option[User]] = {
    OptionT.fromOption(user.id).semiflatMap(id => UserSQL.update(user, id).run.transact(xa).map(_ => user)).value
  }
```

src/main/scala/io/github/pauljamescleary/petstore/infrastructure/repository/doobie/DoobiePetRepositoryInterpreter.scala
```diff
@@ -68,9 +74,14 @@ class DoobiePetRepositoryInterpreter[F[_]: Monad](val xa: Transactor[F])
...
+    def doUpdate(intoId: Long): ConnectionIO[Pet] = PetSQL.update(pet, intoId).run.map(_ => pet)
+    pet.id.traverse(doUpdate).transact(xa)
+  }
+
   def get(id: Long): F[Option[Pet]] = select(id).option.transact(xa)
 
   def delete(id: Long): F[Option[Pet]] = OptionT(get(id)).semiflatMap(pet =>
```
pauljamescleary: I was able to change out the `semiflatMap` in the delete for the following.  Is this ok?

```
  def delete(id: Long): F[Option[Pet]] = {
    for {
      p <- get(id)
      _ <- PetSQL.delete(id).run.transact(xa)
    } yield p
  }
```

src/main/scala/io/github/pauljamescleary/petstore/infrastructure/repository/doobie/DoobiePetRepositoryInterpreter.scala
```diff
@@ -68,9 +74,14 @@ class DoobiePetRepositoryInterpreter[F[_]: Monad](val xa: Transactor[F])
...
 
-  def put(pet: Pet): F[Pet] =
+  def create(pet: Pet): F[Pet] =
     insert(pet).withUniqueGeneratedKeys[Long]("ID").map(id => pet.copy(id = id.some)).transact(xa)
 
+  override def update(pet: Pet): F[Option[Pet]] = {
+    def doUpdate(intoId: Long): ConnectionIO[Pet] = PetSQL.update(pet, intoId).run.map(_ => pet)
+    pet.id.traverse(doUpdate).transact(xa)
+  }
+
```
pauljamescleary: could this update instead be...

```
    OptionT.fromOption[F](pet.id).semiflatMap { id =>
      PetSQL.update(pet, id).run.transact(xa).map(_ => pet)
    }.value
```

src/main/scala/io/github/pauljamescleary/petstore/infrastructure/repository/doobie/DoobieUserRepositoryInterpreter.scala
zakpatterson: Yes I think you and I were fiddling with the same thing for the last few minutes. We do have to include the explicit type parameter `[F]` there.

zakpatterson: I came up with this:

```
  def update(user: User): F[Option[User]] = OptionT.fromOption[ConnectionIO](user.id).semiflatMap { id =>
    UserSQL.update(user, id).run
  }.value.map(_ => user.some).transact(xa)
```

For some reason I think it's nice to keep the transact call at the end.

zakpatterson:
yeah i agree, I have no idea what `semiflatMap` means, I was just looking for a type signature that fit what I needed.

pauljamescleary:
I totally get the usefulness of the function though, it makes so much sense.  Cats has so many goodies it just takes time to find them all.  Just kinda need a comment on it, and make sure it is consistent.

I really like your introduction of `OptionT` as it makes all too much sense for lots of code paths in the app.

src/main/scala/io/github/pauljamescleary/petstore/infrastructure/repository/doobie/DoobieUserRepositoryInterpreter.scala
pauljamescleary: word.  Can we use `as` instead of `map` here?

zakpatterson: Yeah, I'm going to rebase on top of the one you merged, and include that then

src/main/scala/io/github/pauljamescleary/petstore/infrastructure/repository/doobie/DoobiePetRepositoryInterpreter.scala
zakpatterson: yeah I think if a method provided by cats fits a use case we have very well, then we should not shy from using it.  The use case itself helps to instruct people of its use.

src/main/scala/io/github/pauljamescleary/petstore/infrastructure/repository/doobie/DoobiePetRepositoryInterpreter.scala
zakpatterson:
```
OptionT.fromOption[ConnectionIO](pet.id).semiflatMap(id =>
    PetSQL.update(pet, id).run.as(pet)
  ).value.transact(xa)
```
Same thing, adding this

src/main/scala/io/github/pauljamescleary/petstore/infrastructure/repository/doobie/DoobiePetRepositoryInterpreter.scala
zakpatterson: There are a bunch of these in files not touched by this PR. I'll track them all down and amend in a separate PR if that's ok

zakpatterson:
Rebased down to three commits

zakpatterson:
Only thing here I'm not sure about is how to document OptionT usage. Of course the user can go ahead and look up how cats monad transformers work. I think adding a bunch of contextualization within the source isn't great. Maybe documenting that is partly commit messages and partly forthcoming tut documentation?

src/main/scala/io/github/pauljamescleary/petstore/infrastructure/repository/doobie/DoobieUserRepositoryInterpreter.scala
```diff
@@ -43,16 +48,20 @@ class DoobieUserRepositoryInterpreter[F[_]: Monad](val xa: Transactor[F])
 
   import UserSQL._
 
-  def put(user: User): F[User] =
+  def create(user: User): F[User] =
     insert(user).withUniqueGeneratedKeys[Long]("ID").map(id => user.copy(id = id.some)).transact(xa)
 
+  def update(user: User): F[Option[User]] = OptionT.fromOption[ConnectionIO](user.id).semiflatMap { id =>
+    UserSQL.update(user, id).run
+  }.value.map(_ => user.some).transact(xa)
```
pauljamescleary: can we do `.as` instead of `.map(_ =>)`

pauljamescleary:
Looks like a rebase is needed.  Also, just one small comment if we can use `as` and this is good to go.

pauljamescleary:
Yea, let's not worry about comments in the code yet.  With the standard pattern applied, it takes less for someone to grok what is happening.

I have a few more items to take care of and I will create a documentation site.  Those items are:

- add validations for constructing domain objects
- get authentication working, update the database so things are attached to users

Well, that's it really.  Hoping someone can help out with ScalaJS at somepoint as I don't work with it, that would complete this project.

zakpatterson:
Great - I should be able to take a stab at the auth stuff in a couple days. I just implemented it on another project with this stack so it would be nice to try it again and improve on it.

pauljamescleary:
:1st_place_medal:

PR #67, opened 2018-01-29
Remove unused imports.

sergey-lagutin:


pauljamescleary:
Curious as to why these didn't fail the compiler :/

Thanks!

PR #68, opened 2018-01-29
Use .as instead of .map to replace inner value in a functor

zakpatterson:
This is a common enough operation that we should use what cats
gives us, `.as` instead of `.map` for the use `.as(v)` instead of
`.map(_ => v)`


pauljamescleary:
The more I learn about cats, the more awesome it is.

Issue #74, opened 2018-02-02
Update endpoints should be _PUT /<entity>/<entity key>_


pauljamescleary:
This applies to both the updatePet endoint as well as updateUser.

pauljamescleary:
References #20

pauljamescleary:
Closed by #75

PR #75, opened 2018-02-03
Update put endpoints to take key where the update should occur

pauljamescleary:
Previously, we were running PUT to /users

In order to update a user, we should PUT to /users/<userName>

Modify the put pet endpoint to put onto the pet id as well

### DIFF
--- a/functional_test/pet_store_client.py
+++ b/functional_test/pet_store_client.py
@@ -40,7 +40,7 @@ class PetStoreClient(object):
         :param pet:
         :return:
         """
-        url = urljoin(self.index_url, '/pets')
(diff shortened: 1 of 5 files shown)


pauljamescleary:
Not 100% on this.  I looked at https://github.com/spring-projects/spring-petclinic/blob/6730a229a0cf22562fe148d68c3f33d03c44e539/src/main/java/org/springframework/samples/petclinic/owner/OwnerController.java#L111

And they simply overwrote the incoming entity with the userName from the URI path.

This makes sense, as the userName is unique.  We do not allow people to change user names however.

The alternative is more work.  tbh, not sure it is worth the complexity

pauljamescleary:
Closes #74

Issue #49, opened 2017-12-24
Add authentication middleware


pauljamescleary:
Implement authentication middleware for logged in endpoints.

It appears as though TSec is the way to go here...

https://jmcardon.github.io/tsec/docs/http4s/auth-jwt.html

Need a few things here:

1. Need some kind of login endpoint.  That endpoint will take the user and password and 
generate the token
2. Need authentication middleware using TSec
3. Need to update any HttpServices that should be authenticated.  Certainly placing orders

pauljamescleary:
This issue introduces the concept of a User.  Authenticated endpoints (which include all endpoints except signup and login) will now need to be Auth'd.

All services will have to take a user.  For example, if someone POSTs a new pet, we should associate the user id of the user who created it.  Similarly, when someone POSTS a new order, we should save the user id on the order as well.

I am still working through the particulars of TSec, may reach out for some help.

monadplus:
@paulcleary @zakpatterson I wanted to give a try to `TSec` so I started working on the issue. I have the implementation almost ready but I python is a no-no for me. Do you want me to leave PR without the Python part for someone who is willing to help with it ? Or nah ?

zakpatterson:
I think starting a work-in-progress PR for the TSec stuff, and just leaving off the integration testing for it is a great idea. 

As an aside, I do think the python has instructive value for people that come to the project from a python background. I also think adding integration testing using http4s's client libraries would be valuable. Once integration testing is built out using http4s I think there would be a case to be made to drop the python one since maintaining both would be annoying.

monadplus:
I agree. It would be great for someone new to fs2 to build integration tests using the streaming http client from http4s. Although you could just run the non-streaming one which is designed specifically for testing.

I'll try to finish and polish the work-in-progress PR about TSec during the next week. Some feedback will be nice :)

PR #79, opened 2018-02-05
[**WIP] Add tsec based authentication and password hashing

zakpatterson:
Issue #49 

Password hashing is modeled as a separate algebra, parameterized by
the hashing strategy to be used (BCrypt, SCrypt etc) 

Authentication is built into the UserRepositoryAlgebra

* Needs functional testing
* Still not checking user authentication before providing data


src/main/scala/io/github/pauljamescleary/petstore/domain/users/UserRepositoryAlgebra.scala
```diff
@@ -1,19 +1,14 @@
 package io.github.pauljamescleary.petstore.domain.users
 
-import scala.language.higherKinds
+import cats.data.OptionT
+import tsec.authentication.BackingStore
 
-trait UserRepositoryAlgebra[F[_]] {
-  def create(user: User): F[User]
+trait UserBackingStore[F[_]] extends BackingStore[F, Long, User]
```
pauljamescleary: does it make sense to have the repo extend the backing store, or have the UserRepository accept the backing store as a constructor param?  "has a" vs. "is a"

src/main/scala/io/github/pauljamescleary/petstore/domain/users/UserService.scala
```diff
@@ -1,42 +1,39 @@
...
     for {
       _ <- validation.doesNotExist(user)
-      saved <- EitherT.liftF(userRepo.create(user))
+      saved <- EitherT.liftF(userRepo.put(user))
     } yield saved
 
-  def getUser(userId: Long)(implicit M: Monad[F]): EitherT[F, UserNotFoundError.type, User] =
-    EitherT.fromOptionF(userRepo.get(userId), UserNotFoundError)
+  def getUser(userId: Long): EitherT[F, UserNotFoundError.type, User] =
+    EitherT.fromOptionF(userRepo.get(userId).value, UserNotFoundError)
```
pauljamescleary: These `.value`s everywhere are unsettling, curious as to why these were needed.

src/main/scala/io/github/pauljamescleary/petstore/infrastructure/authentication/PasswordHasherCryptInterpreter.scala
```diff
@@ -0,0 +1,17 @@
+package io.github.pauljamescleary.petstore
+package infrastructure.authentication
+
+import cats.effect.Sync
+import tsec.passwordhashers.PasswordHash
+import domain.authentication.CryptAlgebra
+import tsec.passwordhashers.core.PasswordHasher
+
+class PasswordHasherCryptInterpreter[F[_] : Sync, A](
+  implicit H: PasswordHasher[A]
```
pauljamescleary: is there a reason to have the `PasswordHasher` be implicit?

src/main/scala/io/github/pauljamescleary/petstore/infrastructure/endpoint/UserEndpoints.scala
```diff
@@ -33,7 +64,7 @@ class UserEndpoints[F[_]: Effect] extends Http4sDsl[F] {
         }
     }
 
-  private def updateEndpoint(userService: UserService[F]): HttpService[F] =
+  private def updateEndpoint(userService: UserService[F], Auth: AuthService): HttpService[F] =
```
pauljamescleary: I take it that all these endpoints actually have to do something with `Auth: AuthService` at some point?

src/main/scala/io/github/pauljamescleary/petstore/infrastructure/repository/doobie/DoobieUserRepositoryInterpreter.scala
```diff
@@ -11,24 +10,24 @@ import SQLPagination._
 
 private object UserSQL {
   def insert(user: User): Update0 = sql"""
-    INSERT INTO USERS (USER_NAME, FIRST_NAME, LAST_NAME, EMAIL, PASSWORD, PHONE)
-    VALUES (${user.userName}, ${user.firstName}, ${user.lastName}, ${user.email}, ${user.password}, ${user.phone})
+    INSERT INTO USERS (USER_NAME, FIRST_NAME, LAST_NAME, EMAIL, HASH, PHONE)
```
pauljamescleary: Guessing the flyway migration would also need to be modified

zakpatterson: This time I added a second migration, so that commits prior to this one would still work.

src/main/scala/io/github/pauljamescleary/petstore/domain/users/UserRepositoryAlgebra.scala
zakpatterson: The backing store is a generic subset of the UserRepositoryAlgebra I think, seems like our algebra is an extension of it. 
```
  trait BackingStore[F[_], I, V] {
    def put(elem: V): F[V]
    def get(id: I): OptionT[F, V]
    def update(v: V): F[V]
    def delete(id: I): F[Unit]
  }
```

src/main/resources/db/migration/V2__FixPasswordsToBeHashedSecure.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE USERS ALTER COLUMN PASSWORD RENAME TO HASH;
```
zakpatterson: Migration here. This is a trivial change, initially I thought we would have to create another table to store token expiry, but as I understand the stateless JWT authenticator can do it cryptographically without a store for the tokens.

src/main/scala/io/github/pauljamescleary/petstore/infrastructure/authentication/PasswordHasherCryptInterpreter.scala
pauljamescleary: Got it, ok

src/main/scala/io/github/pauljamescleary/petstore/infrastructure/endpoint/UserEndpoints.scala
```diff
@@ -1,28 +1,59 @@
...
+          name = login.userName
+          user <- userService.getUserByName(name).leftMap(_ => UserAuthenticationFailedError(name))
+          valid <- EitherT.liftF(crypt.check(login.password, PasswordHash[A](user.hash)))
+          resp <- EitherT.liftF(Ok())
+        } yield resp
+
+        // Handle error cases:
+        action.value.flatMap{
+          case Right(resp) => resp.pure[F]
+          case Left(UserAuthenticationFailedError(name)) => Conflict(s"Authentication failed for user $name")
```
pauljamescleary: Is `Conflict` the correct response?

zakpatterson: No. I think 401 is right, but I couldn't figure out how to instantiate it, the `apply` on `Unauthorized` is confusing, since a 401 requires some WWW-Authenticate header. Maybe just `400`/`BadRequest`

src/main/scala/io/github/pauljamescleary/petstore/infrastructure/authentication/PasswordHasherCryptInterpreter.scala
zakpatterson: Yeah i think this kind of works, because you can just do `new PasswordHasherCryptInterpreter[IO, SCrypt]` to specify which alg you want.

zakpatterson:
Hit a bit of a wall here. This tiny patch here causes all the tests to fail: 

https://github.com/zakpatterson/scala-pet-store/commit/dfb7951954a324b3cafacbe557889f0a14dfe3f6

pauljamescleary:
A few more comments since taking a look at this...

I don't think the crypt / authentication / LoginRequest stuff lives in the domain.  If we are exclusively using it in the endpoints, then all of that stuff should live in infrastructure.

I don't think we need to tie the user repository algebra to the tsec BackingStore in the domain.  Seems like that can be something that we do in the interpreter or keep it separate.  It muddies the definition of the user repository algebra which is supposed to be domain centric.

This is great stuff.  Seems like you are stopped right now on actually using the auth stuff?  I suppose that in order to actually do the auth stuff, we would need to have a stateful client in python and first login, and then second keep the JWT around for each request?  Treading new water here so a little unclear.

zakpatterson:
> I don't think we need to tie the user repository algebra to the tsec BackingStore in the domain. Seems like that can be something that we do in the interpreter or keep it separate. It muddies the definition of the user repository algebra which is supposed to be domain centric.

This makes sense, hard to balance soc and code reuse in this case.  

> I suppose that in order to actually do the auth stuff, we would need to have a stateful client in python and first login, and then second keep the JWT around for each request? 

My understanding is that the python requests session already handles this, or should handle this. I might need some help on the pythong testing side.  

I've also learned a bit about the client side of http4s. I agree with your rationale in the readme about the python client, but i think it might also be valuable to show more testing with http4s. I'll open another issue about that after this.

src/main/scala/io/github/pauljamescleary/petstore/infrastructure/authentication/UserBackingStore.scala
```diff
@@ -0,0 +1,23 @@
...
+
+abstract class UserBackingStore[F[_]] extends BackingStore[F, Long, User]
+
+object UserBackingStore {
+
+  /**
+    * A transformation between the UserRepositoryAlgebra in our domain and a BackingStore
+    * for users which is required by tsec at our endpoint.
+    */
+  def trans[F[_]](A: UserRepositoryAlgebra[F]): UserBackingStore[F] = new UserBackingStore[F] {
```
zakpatterson: I think this could be named a little better, but can't come up with anything right now. I'm using it as `UserBackingStore.trans(userAlgebra)`, so maybe that's good.

I think this addresses the code-reuse vs SoC dilemma ok, let me know what you think.

functional_test/pet_store_client.py
```diff
@@ -132,6 +132,18 @@ def signup_user(self, user):
...
+        """
+        Logs in an existing user
+        """
+        url = urljoin(self.index_url, '/login')
+
+        response = self.session.post(url,
+                                     headers=self.headers,
+                                     data=json.dumps(user))
+
+        return response
```
zakpatterson: I attempted to use 
```
        self.headers = dict(self.headers, **response.headers)
```
Here, but just sending that header on each subsequent request does not work. The response headers after login look like:

```
{'Date': 'Tue, 06 Feb 2018 22:33:25 GMT', 'Content-Length': '0', 'Authorization': 'Bearer eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzdWIiOiIxIiwiZXhwIjoxNTE3OTU3MDA1LCJpYXQiOjE1MTc5NTY0MDUsImp0aSI6IjYyNjAxYzljNzczOThmODc1M2ExMDYzMzFkNDBkMjViMWFlYjQ4ZDUyZjNiNjFlMjFkY2E3NzI5YjdiMzY1ODUifQ.vAJEogj6TtCkwU1w4ZGU4Jz812YFZKsM4CaBp6RZErU'}
```

PR #88, opened 2018-07-20
Update post-pet.sh

cnguy:
Small update to URLs in the scripts; App uses standard REST API URLs instead of query parameters.


pauljamescleary:
Thanks for the update!

Issue #86, opened 2018-05-19
Improve enums


pauljamescleary:
Currently, the enums we use are unsavory.  

Let's use Enumeratum (and circe Enumeratum).  Should be a seemless add library dependencies, imports where appropriate, and refactor the few enums that we have.

PR #89, opened 2018-08-06
Add enumeratum

ml10:
This should address issues #86. 

I didn't change `ValidationError.scala` as I don't think it would fit well with Enumeratum, given the mix of `case class`es and `case object`s.


src/main/scala/io/github/pauljamescleary/petstore/domain/pets/PetStatus.scala
```diff
@@ -1,20 +1,12 @@
...
+  case object Available extends PetStatus
+  case object Pending extends PetStatus
+  case object Adopted extends PetStatus
 
-  def nameOf(status: PetStatus): String = status match {
-    case Available => "Available"
-    case Pending => "Pending"
-    case Adopted => "Adopted"
-  }
+  val values = findValues
```
pauljamescleary: what is the reason for `val values = ...`?  

Also, perhaps add an explicit type for `values`?

ml10: I'm no expert in enumeratum, but I believe is boilerplate required for the macros to work. See [this comment in the code](https://github.com/lloydmeta/enumeratum/blob/master/enumeratum-core/src/main/scala/enumeratum/Enum.scala#L150-L156). There's [an issue around dealing with the boilerplate](https://github.com/lloydmeta/enumeratum/issues/141), but it's still open.

As for the adding in an explicit type to `values`, it's a `scala.collection.immutable.IndexedSeq[A]`. Would you prefer to be as explicit as possible or would you rather a plain `Seq[A]`?

pauljamescleary: @ml10 the `values` doesn't have an explicit return type in the Enumeratum docs (just looked) so it is ok to leave it out.

pauljamescleary: @martin-g  👍  ?

PR #93, opened 2018-08-31
WIP Frontend scalajs

RobertLemmens:
Ive been using this project as a template for a while, and seeing there's no frontend yet, i thought i'd go ahead and demonstrate one way to possibly do it, maybe get the ball rolling! This PR is meant to get people to start thinking about possible implementations and is not a finished frontend.

I ordered it as a multimodule project and have the backend project actually serve the .html(added a simple endpoint for this). The code from frontend gets compiled to JS and added to the backend when compiling the backend, so a simple reStart on the backend project is enough, thats pretty nice! Maybe a root build task in build.sbt can make this even easier. The frontend can exist on its own perfectly fine as well. This way however you can create a fat jar containing front and back. 

What are your thoughts on a multi module project? Favor more decoupled front and back? Open discussion 😄 !


pauljamescleary:
Thanks @RobertLemmens \o/

For your questions, I am not sure what would be considered best practice for scalajs projects.  So I would have to defer.  From an operational perspective, it is easier to have a single executable.  With Http4s authentication, you can have multiple modes of authentication, so we could for example use OAUTH for pure API authentication, and JWT for the scalajs front end.

What has your experience been wrt scalajs in terms of organization (given I have none ;) ?

RobertLemmens:
My experience with scalaJS has been somewhat mixed. It does a good job at what it sets out to do but Ive noticed i spent more time writing facades for JS stuff i need rather than writing real domain. Once you have your facades its pretty much the same as always :). The nicest thing about it is ofcourse the possibility of shared code, which is why i use it now and then. But this is a design choice that might not always fit the requirements. When appropriate, i'd opt for a standalone frontend based on whatever JS framework. But this might be out of scope for this project since it would neither be scala nor based on typelevel or functional best practices. Based on that i believe scalaJS (or one of its diverts) is the best fit for the pet store.

pauljamescleary:
@RobertLemmens thinking out loud, part of the value of ScalaJS is to allow reusing the same models between front-end and back-end.  Is that represented in your re-organization?

I would think that is easier if everything was the same project, or there was a "core" that is reused between backend and frontend.

RobertLemmens:
@pauljamescleary Adding a core module is in line with how it's setup now. I will add a core module later today in my repo, see how it looks.

RobertLemmens:
@pauljamescleary update the repo with shared sources. I removed only the models from the backend project into the shared sources folder. Tested running it in sbt with reStart on the backend project and seems to work as expected (haven't updated the build.sh yet). Called it shared instead of core as per https://www.scala-js.org/doc/project/cross-build.html , but anything works ;). Take a look when u got the time.

pauljamescleary:
@RobertLemmens I guess I take back what I said.  I saw an example from Christopher Davenport and another one that split the project to frontend and backend.  Seems like that maybe the convention.

Here is a good example - https://github.com/eugengarkusha/http4s-scalajs
That one uses tsec for auth.  Would love to explore that a bit.

Perhaps we create a branch for this work and playground on it?  Could also tinker on this PR.  Let me know your thoughts.

RobertLemmens:
@pauljamescleary sounds good. That example has the package structure the same as my branch. I'll have to take a look at how tsec is used there, I'm not too experienced on that, but it looks good. If you can get a new branch or my branch in Ill start contributing there, work on package structure and get a frontend going.

PR #96, opened 2018-10-09
WIP: Experimental using tsec for authentication

wibisono:
An attempt to add JWT authentication, in essence what I am doing is following blindly example from:
- https://jmcardon.github.io/tsec/docs/http4s/auth-jwt.html
- https://gitter.im/tsecc/Lobby?at=5b83999c60f9ee7aa4986b61

Things I am not sure of: 
- Is it true that I need to use UserService as identity backingStorage for JWTAuthenticator?
- Whether or not I need the token store as done now in AuthenticationService I introduced, this is blind example following.

I added dummy secure GET end point in user endpoints, to give an example how an endpoint requiring authentication looks like, but I did not touch the rest of the endpoint, with some test on how it behave when user accessed it with proper headers after logging (auth case) and when it does not provide required Auth headers.

I have no idea how to update functional tests since I can't really parse the python code :)

### DIFF
--- a/src/main/scala/io/github/pauljamescleary/petstore/Server.scala
+++ b/src/main/scala/io/github/pauljamescleary/petstore/Server.scala
@@ -1,14 +1,15 @@
 package io.github.pauljamescleary.petstore
 
-import config.{DatabaseConfig, PetStoreConfig}
-import domain.users._
-import domain.orders._
-import domain.pets._
-import infrastructure.endpoint.{OrderEndpoints, PetEndpoints, UserEndpoints}
-import infrastructure.repository.doobie.{DoobieOrderRepositoryInterpreter, DoobiePetRepositoryInterpreter, DoobieUserRepositoryInterpreter}
 import cats.effect._
 import fs2.StreamApp.ExitCode
 import fs2.{Stream, StreamApp}
+import io.github.pauljamescleary.petstore.config.{DatabaseConfig, PetStoreConfig}
+import io.github.pauljamescleary.petstore.domain.authentication.AuthenticationService
+import io.github.pauljamescleary.petstore.domain.orders._
+import io.github.pauljamescleary.petstore.domain.pets._
+import io.github.pauljamescleary.petstore.domain.users._
+import io.github.pauljamescleary.petstore.infrastructure.endpoint.{OrderEndpoints, PetEndpoints, UserEndpoints}
+import io.github.pauljamescleary.petstore.infrastructure.repository.doobie.{DoobieOrderRepositoryInterpreter, DoobiePetRepositoryInterpreter, DoobieUserRepositoryInterpreter}
 import org.http4s.server.blaze.BlazeBuilder
 import tsec.mac.jca.HMACSHA256
 import tsec.passwordhashers.jca.BCrypt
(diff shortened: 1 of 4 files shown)


wibisono:
Somehow @zakpatterson's PR makes more sense :D I am on the verge of closing and forgetting about this.

pauljamescleary:
Thanks @wibisono , I don't have a ton of experience on the auth side so admittedly this is one for experimentation.

I think the end state would be:

1. Support dual means of authentication (oauth2 and jwt)
1. A scala js application could use jwt / secure cookies to interact with the backend from a web application
1. Clients could directly access the backend using OAUTH2 (or similar).

Http4s does support multiple auth mechanisms.

tbf, I haven't revisited this in a while.  I know TSEC was under active development a while back and was changing too frequently for me to keep up with.

I also haven't dug into the scalajs app.  #93 exists as a starting point for a scala js application.  I would prefer to have everything in the same project.

pauljamescleary:
I might see if @jmcardon or @rossabaker are around (may not be)

rossabaker:
I think @jmcardon was trying to give tsec away. There are still PRs as recent as a day old, so there is still interest. Does it have a maintainer?

pauljamescleary:
👋  @rossabaker , not sure about tsec.

Am curious if you had guidance / thoughts on ScalaJS + JWT combined with OATH2 for non UI authentication on http4s.  I would really like to make that happen, but haven't had a ton of time recently.

rossabaker:
Let me ask on the http4s channel. I've (dirty secret) never done auth on http4s.

jmcardon:
@rossabaker not giving it away, just considering archival until I could maintain it again. I found maintainers, so it's not going anywhere.

@paulcleary im not really supporting scalajs atm... So that may be an issue.

jmcardon:
w.r.t tsec:
- It is a lot more stable now, since 0.0.1-RC1 is out, no major changes will be made. Same wiht 0.1 of tsec.
- OAuth support is really young but promising.
- I'm not entirely out of the picture but my time _is_ quite limited, so apologies if I cannot get to this for a while. I hope that with the more stable tsec, we've put out enough tools to be able to implement most things, if, at worst, from tsec-derived primitives.

pauljamescleary:
Thanks @rossabaker and @jmcardon for chiming in!

I would love to put together a plan for this "major final piece" for the pet store.  Having a scalajs front end (maybe using Outwatch?), with secure cookie / JWT for authentication, as well as a direct REST api authentication would be the bees knees.

A lot of people have the problem (including myself) of needing both a UI as well as a REST api to be consumable by clients.

Any guidance, or people / resources I can check out for building this into the pet store would be much appreciated.

pauljamescleary:
Also, found this with some googling might give some guidance? - https://github.com/eugengarkusha/http4s-scalajs

wibisono:
> Also, found this with some googling might give some guidance? - https://github.com/eugengarkusha/http4s-scalajs

Scalajs is still foreign to me also, so I guess we'll wait for other scalajs native volunteer to do this :)

pauljamescleary:
@wibisono me too.  We have some folks over here doing some stuff with it.  @RobertLemmons also has a PR up to start that effort.

PR #102, opened 2018-11-25
Fix flyway migration to remove doobie transactor argument

zakpatterson:
Flyway initiates its own JDBC connection to the database, and then
closes it when the migration is done. There's no reason to require
a HikariTransactor in test just to use its JDBC datasource in order to apply
migrations via Flyway, so we can just create a JDBC datasource from
our configuration and pass it to flyway.

### DIFF
--- a/src/main/scala/io/github/pauljamescleary/petstore/Server.scala
+++ b/src/main/scala/io/github/pauljamescleary/petstore/Server.scala
@@ -18,38 +18,34 @@ import scala.concurrent.ExecutionContext.Implicits.global
 object Server extends IOApp {
   private val keyGen = HMACSHA256
 
-  def createServer[F[_] : ContextShift : ConcurrentEffect : Timer]: F[ExitCode] =
+  def createServer[F[_] : ContextShift : ConcurrentEffect : Timer]: Resource[F, ExitCode] =
     for {
-      conf           <- PetStoreConfig.load[F]
-      signingKey     <- keyGen.generateKey[F]
-      xar            =  DatabaseConfig.dbTransactor(conf.db, global, global)
-      exitCode       <- xar.use { xa =>
-        val petRepo        =  DoobiePetRepositoryInterpreter[F](xa)
-        val orderRepo      =  DoobieOrderRepositoryInterpreter[F](xa)
-        val userRepo       =  DoobieUserRepositoryInterpreter[F](xa)
-        val petValidation  =  PetValidationInterpreter[F](petRepo)
-        val petService     =  PetService[F](petRepo, petValidation)
-        val userValidation =  UserValidationInterpreter[F](userRepo)
-        val orderService   =  OrderService[F](orderRepo)
-        val userService    =  UserService[F](userRepo, userValidation)
-        val services       =  PetEndpoints.endpoints[F](petService) <+>
-                              OrderEndpoints.endpoints[F](orderService) <+>
-                              UserEndpoints.endpoints[F, BCrypt](userService, BCrypt.syncPasswordHasher[F])
-        val httpApp = Router("/" -> services).orNotFound
-
-        val init : F[Unit] = xa.configure(DatabaseConfig.initializeDb(_))
-
-        val build : F[ExitCode] = BlazeServerBuilder[F]
-          .bindHttp(8080, "localhost")
-          .withHttpApp(httpApp)
-          .serve
-          .compile
-          .drain
-          .as(ExitCode.Success)
-
-        init *> build
-      }
+      conf           <- Resource.liftF(PetStoreConfig.load[F])
+      signingKey     <- Resource.liftF(keyGen.generateKey[F])
+      xa             <- DatabaseConfig.dbTransactor(conf.db, global, global)
+      petRepo        =  DoobiePetRepositoryInterpreter[F](xa)
+      orderRepo      =  DoobieOrderRepositoryInterpreter[F](xa)
+      userRepo       =  DoobieUserRepositoryInterpreter[F](xa)
+      petValidation  =  PetValidationInterpreter[F](petRepo)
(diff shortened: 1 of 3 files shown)


zakpatterson:
Oops, sorry about out of date local branch. will revise

src/main/scala/io/github/pauljamescleary/petstore/config/DatabaseConfig.scala
```diff
@@ -17,16 +17,17 @@ object DatabaseConfig {
...
+  def initializeDb[F[_]](dbConfig: DatabaseConfig)(implicit S: Sync[F]): F[Unit] =
     if (dbConfig.url.contains(":h2:")) {
-      xa.configure { ds =>
-        S.delay {
-          val fw = new Flyway()
-          fw.setDataSource(ds)
-          fw.migrate()
-          ()
-        }
+      S.delay {
```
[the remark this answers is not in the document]
zakpatterson: I hadn't considered the if statement, but I think you're right. Do we need the if statement if we're not requiring this project work with multiple database systems? If we're just using H2, then I don't see the reason to check or have it as part of the configuration.

I'm not sure I get the transactor question. The migration is applied in a sequential operation handled by whatever context is managed by `F`. So it doesn't need a transactor to exist before this function is called. We could just provide another method which takes a `Datasource` as its input?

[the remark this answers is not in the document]
pauljamescleary: Right, we just need a DataSource to be provided in here.  Would prefer that to be an argument as opposed to hard-coding to h2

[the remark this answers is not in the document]
pauljamescleary: I think we drop the `if`, and pass in a `DataSource` that we can use with `Flyway`.

pauljamescleary:
@zakpatterson looks like merging the other one broke this one.  Sorry if I merged out of order.  Seems like a few conflicts to be resolved.

src/main/scala/io/github/pauljamescleary/petstore/Server.scala
```diff
@@ -18,34 +18,38 @@ import scala.concurrent.ExecutionContext.Implicits.global
 object Server extends IOApp {
   private val keyGen = HMACSHA256
 
-  def createStream[F[_] : ContextShift : ConcurrentEffect : Timer]: F[ExitCode] =
+  def createServer[F[_] : ContextShift : ConcurrentEffect : Timer]: F[ExitCode] =
     for {
-      conf           <- PetStoreConfig.load[F]
```
zakpatterson: sorry this diff is dirty. Fixing.

zakpatterson:
Ok, I rebased and fixed up this commit, should be good now.

src/main/scala/io/github/pauljamescleary/petstore/config/DatabaseConfig.scala
```diff
@@ -18,23 +19,12 @@ object DatabaseConfig {
...
-        fw.migrate()
-        ()
-      }
-    } else {
-      S.pure(())
+  def initializeDb[F[_]](ds: DataSource)(implicit S: Sync[F]): F[Unit] =
+    S.delay {
+      val fw = new Flyway()
+      fw.setDataSource(ds)
+      fw.migrate()
```
pauljamescleary: ok, looks like we can `fw.setDataSource(url, user, pass)` instead of passing the DataSource.

We can avoid hard-coding to h2 if we pass in the `DatabaseConfig`

src/test/scala/io/github/pauljamescleary/petstore/infrastructure/repository/doobie/package.scala
```diff
@@ -16,12 +17,28 @@ package object doobie {
       cfg.password           // password
     )
 
+  /*
+   * Create an h2 DataSource which flyway can use to initialize the database.
```
zakpatterson: We're considering whether it makes sense to hardcode H2 in test here.

src/main/scala/io/github/pauljamescleary/petstore/config/DatabaseConfig.scala
zakpatterson: That does work actually, good catch. Even after reading the flyway source I'm not sure how it finds the correct driver based on the URL, but it does seem to work well.

src/test/scala/io/github/pauljamescleary/petstore/infrastructure/repository/doobie/package.scala
zakpatterson: And apparently it does not because flyway does actually autoselect the correct driver given just the URL somehow.

zakpatterson: Ah, there it is: https://github.com/flyway/flyway/blob/0260b5c37eb0c8f3bbc331c1e293710e32324deb/flyway-core/src/main/java/org/flywaydb/core/internal/jdbc/DriverDataSource.java#L280

src/main/scala/io/github/pauljamescleary/petstore/config/DatabaseConfig.scala
```diff
@@ -20,10 +19,11 @@ object DatabaseConfig {
   /**
     * Runs the flyway migrations against the target database
     */
-  def initializeDb[F[_]](ds: DataSource)(implicit S: Sync[F]): F[Unit] =
+  def
+  initializeDb[F[_]](cfg : DatabaseConfig)(implicit S: Sync[F]): F[Unit] =
```
zakpatterson: oops, fixed this

pauljamescleary:
Cool, thanks @zakpatterson !!!

PR #104, opened 2018-12-05
make PageSizeMatcher and OffsetMatcher optional

Phill101:
pageable routes without any query parameters should return response with default offset and page size

### DIFF
--- a/functional_test/pet_store_client.py
+++ b/functional_test/pet_store_client.py
@@ -9,6 +9,17 @@ from urlparse import urlparse
 from urlparse import parse_qs
 from urlparse import urlsplit
 
+def map_to_query_params(m, *args):
+    mm = list(filter(lambda k: k[0] in args, m.items()))
+    if len(mm) > 0:
+        return "?" + "&".join(["{0}={1}".format(k, v) for k,v in mm])
+    else:
+        return ""
+
+def paginate(m):
+    return map_to_query_params(m, "pageSize", "offset")
+
+
 class PetStoreClient(object):
     def __init__(self, url='http://localhost:8080'):
         self.index_url = url
@@ -54,12 +65,12 @@ class PetStoreClient(object):
 
         return self.session.request('GET', url, self.headers)
 
-    def list_pets(self, page_size=10, offset=0):
+    def list_pets(self, **kwargs):
         """
-        Returns a list of pets
+        Returns a list of pets, taking optional keyword argument page_size and offset for pagination
         :return:
         """
-        url = urljoin(self.index_url, "/pets?pageSize={0}&offset={1}".format(page_size, offset))
+        url = urljoin(self.index_url, "/pets" + paginate(kwargs))
 
         return self.session.request('GET', url, self.headers)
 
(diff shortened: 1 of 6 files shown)


src/main/scala/io/github/pauljamescleary/petstore/infrastructure/endpoint/UserEndpoints.scala
```diff
@@ -79,10 +79,10 @@ class UserEndpoints[F[_]: Effect, A] extends Http4sDsl[F] {
 
   private def listEndpoint(userService: UserService[F]): HttpRoutes[F] =
     HttpRoutes.of[F] {
-      case GET -> Root / "users" :? PageSizeMatcher(pageSize) :? OffsetMatcher(offset) =>
+      case GET -> Root / "users" :? OptionalPageSizeMatcher(pageSize) :? OptionalOffsetMatcher(offset) =>
         for {
-          retrived <- userService.list(pageSize, offset)
-          resp <- Ok(retrived.asJson)
+          retrieved <- userService.list(pageSize.getOrElse(10), offset.getOrElse(0))
+          resp <- Ok(retrieved.asJson)
```
zakpatterson: much better spelling.

zakpatterson:
Can we add a new functional test showing the /pets and /users routes work now?

Phill101:
> Can we add a new functional test showing the /pets and /users routes work now?
I'll do it tomorrow

pauljamescleary:
I like it, @zakpatterson

zakpatterson:
I added a commit to this PR at: https://github.com/Phill101/scala-pet-store/pull/1, @Phill101 I think you just have to approve that commit and it will show up here.

functional_test/pet_store_client.py
```diff
@@ -9,6 +9,17 @@
 from urlparse import parse_qs
 from urlparse import urlsplit
 
+def map_to_query_params(m, *args):
```
pauljamescleary: I have not tested this, but it appears as though this kind of mapping is already present in the python `requests` library and it is possible I have been doing it all wrong?

The following is from the requests documentation.  Seemingly you can do `self.session.request(..., params=your_dictionary_here)`.  Note I didn't try it yet

```
        :param params: (optional) Dictionary or bytes to be sent in the query
            string for the :class:`Request`.
```

pauljamescleary: Just confirmed this does work.  I think we should use this method, and perhaps a different PR go back and fix all the other methods...

```python
    def list_pets(self, page_size=10, offset=0):
        """
        Returns a list of pets
        :return:
        """
        url = urljoin(self.index_url, "/pets")
        return self.session.request(method='GET', url=url, headers=self.headers, params={'pageSize': page_size, 'offset': offset})
```

zakpatterson: Hey @Phill101 I added another commit to your PR that does what Paul is talking about here. Just merge that and this should be resolved. https://github.com/Phill101/scala-pet-store/pull/2 

Only problem is the way I've done it the caller of the client method can pass any parameters and they'll be put into the query string. I agree with Paul tho let's revise all these methods for parameter usage in one later PR.

PR #107, opened 2018-12-28
Revise server creation to use http4s's resource instead of stream

zakpatterson:
Thanks to @rossabaker for chat help in understanding:
https://gitter.im/http4s/http4s?at=5c25581d5064a51f8386d93f

* We were using IO.pure inappropriately, see this comment:
https://github.com/typelevel/cats-effect/blob/v1.0.0/core/shared/src/main/scala/cats/effect/IO.scala#L988
* See http4s docs for discussion about resource/stream:
https://http4s.org/v0.20/service/#running-your-service-as-an-app


pauljamescleary:
Thanks @zakpatterson !

PR #109, opened 2018-12-28
Fix sbt run

### LINKED ISSUE #105: `sbt run` not working
When using `sbt run` to start the project, it appears that `pureconfig.loadConfig` cannot find the `reference.conf` file. It complains that the `petstore` key is missing even though it's there.

```
pureconfig.error.ConfigReaderException: Cannot convert configuration to a io.github.pauljamescleary.petstore.config.PetStoreConfig. Failures are:
  at the root:
    - Key not found: 'petstore'.
```

pauljamescleary:
Fixes #105 

For some reason unknown at this time, `ConfigFactory.load()` fails to find `src/main/resources/application.conf` using `sbt run`; although this seems to work fine in revolver `sbt reStart`

* update `build.sbt` to fork in run, this allows the config factory to load properly

**In addition**

1. Updated the version of sbt to latest
1. Updated the config loading to use pureconfig cats-effect

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -16,41 +16,42 @@ val LogbackVersion         = "1.2.3"
 val ScalaCheckVersion      = "1.14.0"
 val ScalaTestVersion       = "3.0.5"
 val FlywayVersion          = "4.2.0"
-val PureConfigVersion      = "0.9.2"
+val PureConfigVersion      = "0.10.1"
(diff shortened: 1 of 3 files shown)


zakpatterson:
Confirmed `sbt run` works for me on this branch. `fork in run` was all that was needed huh? How weird.

Issue #120, opened 2019-01-04
Switch to circe-config


zakpatterson:
For consistency, we should use circe-config instead of pureconfig.

zakpatterson:
I don't think there could be any measurable advantage, it just seems like a more natural choice for this project since we're already using circe. Also c-c doesn't do kebab-case which seemed weird to me. Here's a line where I used it in another project, although we wouldn't need to use all the ApplicativeAsk complexity here:

https://github.com/clovellytech/http4s-modules/blob/master/src/main/scala/h4sm/package.scala#L16

zakpatterson:
I went ahead and did it, let me know what you think.

franklinchou:
This was closed by #123, no?

zakpatterson:
Fixed with #123, thanks @franklinchou !

PR #123, opened 2019-01-09
Remove pureconfig in favor of circe-config

zakpatterson:
Addresses #120


zakpatterson:
hold off on merging, I'm lobbying for this in circe-config: https://github.com/circe/circe-config/pull/29

pauljamescleary:
We should update the README too (anywhere else that uses pureconfig).

zakpatterson:
It looks like the default load PR will get into circe-config soon, so if you agree I think it makes sense to just merge this now and then revise usage once scala-steward picks up a new release from circe-config.

pauljamescleary:
@zakpatterson sounds good

PR #140, opened 2019-02-05
remove redundant dependency

SethTisue:
the redundancy would normally be harmless, I think, but for
some reason over at https://github.com/scala/community-builds/pull/835
it's harmful in the context of the Scala community build


zakpatterson:
Awesome Seth, very happy for this project to be included.

Issue #149, opened 2019-02-28
Application gets blocked on db connection acquisition if more than 10 concurrent clients connect


irumiha:
I'm not sure if it's about this application or this is a property of all http4s / doobie / hikari applications. It seems that something gets stuck when there are more clients connected than available connections in the connection pool.

How to reproduce: 

1. do sbt dist (to remove the sbt environment as a factor)
2. unpack and run the app 
3. run `siege http://localhost:8080/users`

Use top, htop or whatever to monitor system activity. You will notice nothing is happening. After 30 seconds exceptions start popping up:

```
[scala-execution-context-global-33] ERROR o.h.s.service-errors - Error servicing request: GET /users from 127.0.0.1 
java.sql.SQLTransientConnectionException: HikariPool-1 - Connection is not available, request timed out after 30000ms.
	at com.zaxxer.hikari.pool.HikariPool.createTimeoutException(HikariPool.java:676)
	at com.zaxxer.hikari.pool.HikariPool.getConnection(HikariPool.java:190)
	at com.zaxxer.hikari.pool.HikariPool.getConnection(HikariPool.java:155)
	at com.zaxxer.hikari.HikariDataSource.getConnection(HikariDataSource.java:128)
	at doobie.util.transactor$Transactor$fromDataSource$FromDataSourceUnapplied.$anonfun$apply$5(transactor.scala:260)
	at cats.effect.internals.IORunLoop$.cats$effect$internals$IORunLoop$$loop(IORunLoop.scala:87)
	at cats.effect.internals.IORunLoop$.startCancelable(IORunLoop.scala:41)
	at cats.effect.internals.IOBracket$BracketStart.run(IOBracket.scala:86)
	at cats.effect.internals.Trampoline.cats$effect$internals$Trampoline$$immediateLoop(Trampoline.scala:70)
	at cats.effect.internals.Trampoline.startLoop(Trampoline.scala:36)
	at cats.effect.internals.TrampolineEC$JVMTrampoline.super$startLoop(TrampolineEC.scala:93)
	at cats.effect.internals.TrampolineEC$JVMTrampoline.$anonfun$startLoop$1(TrampolineEC.scala:93)
	at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:23)
	at scala.concurrent.BlockContext$.withBlockContext(BlockContext.scala:85)
[...]

PR #150, opened 2019-02-28
Use different thread pool for connections in dbTransactor.

irumiha:
Don't know why but it seems to solve #149.

### DIFF
--- a/src/main/scala/io/github/pauljamescleary/petstore/Server.scala
+++ b/src/main/scala/io/github/pauljamescleary/petstore/Server.scala
@@ -12,6 +12,7 @@ import org.http4s.server.{Server => H4Server, Router}
 import org.http4s.server.blaze.BlazeServerBuilder
 import org.http4s.implicits._
 import tsec.passwordhashers.jca.BCrypt
+import doobie.util.ExecutionContexts
 import scala.concurrent.ExecutionContext.Implicits.global
 import io.circe.config.parser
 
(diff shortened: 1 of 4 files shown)


src/main/scala/io/github/pauljamescleary/petstore/Server.scala
```diff
@@ -12,14 +12,16 @@ import org.http4s.server.{Server => H4Server, Router}
...
+import doobie.util.ExecutionContexts
 import scala.concurrent.ExecutionContext.Implicits.global
 import io.circe.config.parser
 
 object Server extends IOApp {
   def createServer[F[_] : ContextShift : ConcurrentEffect : Timer]: Resource[F, H4Server[F]] =
     for {
       conf           <- Resource.liftF(parser.decodePathF[F, PetStoreConfig]("petstore"))
-      xa             <- DatabaseConfig.dbTransactor(conf.db, global, global)
+      connEc         <- ExecutionContexts.fixedThreadPool[F](10)
```
zakpatterson: This is a great catch. Do you think it's worth adding a configuration option for the number of connection threads, so we don't have that `10` or some other number in there?

Also, we should probably follow the documentation on connections, here, which your code agrees with, but should also fix/understand the use of global on the following line.
https://tpolecat.github.io/doobie/docs/14-Managing-Connections.html#using-a-hikaricp-connection-pool

irumiha: Re: configuration option - yes, I'll amend the PR.

Actually the suggested use of the unbounded thread pool for transaction execution should also go here. If you agree I'll add that too.

zakpatterson: Yep I think fixing up both and adding the configuration option would make this great.

pauljamescleary: I think just upping the `maxConnectionPoolSize` setting on the hikari CP should do the trick.

I do agree that the separate execution context is the right thing as well.

There should be a way to call `Executors.xxx` and get the right sized thread pool.  it should be based by the number of available processors.

zakpatterson:
I think this is the right correction, @paulcleary give a 👍 since you weighed in on the comments above

PR #163, opened 2019-04-18
Authentication using TSec

monadplus:
This PR addresses the following issues:
 - https://github.com/pauljamescleary/scala-pet-store/issues/49
 - https://github.com/pauljamescleary/scala-pet-store/issues/16

__All python end-to-end testing is not done so, sadly, this will block the PR. I'd appreciate if someone takes responsibility of this part.__

Regarding to the PR, I am quite satisfied with the results, however, they are some points that are not addressed:
 - Symmetric Key used to sign JWT is not persisted.
 - Tokens are not removed from db when expired. I assumed TSec would take responsibility of this part (it has all the resources to do so), but it does not. 
 - I tried to type alias all the hard types from TSec. I tried moving the following ones to `Auth.scala` but it didn't work (idkw)
```scala
type AuthService = TSecAuthService[User, AugmentedJWT[Auth, Long], F]
  type AuthEndpoint = PartialFunction[SecuredRequest[F, User, AugmentedJWT[Auth, Long]], F[Response[F]]]
```
 - Composition of endpoints was tricky because TSec, to avoid a non-authenticated user to crawl the list of endpoints, short-circuits to the first non-authenticated endpoint for the user.

It was a very educational experience from me because when I started this PR I hadd 0 experience with doobie, http4s and tsec.


.gitignore
```diff
@@ -6,3 +6,5 @@ project/target
 .virtualenv
 __pycache__
 *.pyc
+.bloop
+.metals
```
pauljamescleary: Please add a new line to get rid of this.

Also, you are using metals?  Haven't even tried it yet (or haven't tried it in a while).  Wonder how it is going.

src/main/scala/io/github/pauljamescleary/petstore/infrastructure/endpoint/OrderEndpoints.scala
```diff
@@ -6,12 +6,19 @@ import io.circe.syntax._
...
 import io.github.pauljamescleary.petstore.domain.orders.{Order, OrderService}
+import io.github.pauljamescleary.petstore.domain.users.User
+import tsec.authentication.{AugmentedJWT, SecuredRequest, SecuredRequestHandler, TSecAuthService, asAuthed}
+import tsec.jwt.algorithms.JWTMacAlgo
+
+class OrderEndpoints[F[_]: Effect, Auth: JWTMacAlgo] extends Http4sDsl[F] {
 
-class OrderEndpoints[F[_]: Effect] extends Http4sDsl[F] {
+  type AuthService = TSecAuthService[User, AugmentedJWT[Auth, Long], F]
+  type AuthEndpoint = PartialFunction[SecuredRequest[F, User, AugmentedJWT[Auth, Long]], F[Response[F]]]
```
pauljamescleary: this is really nice.  Seem to repeat these type aliases in all the endpoints, can we have these reused across all endpoints?

.gitignore
monadplus: I always forget about adding a new line at the end of the files.

I use vscode for haskell and sometimes for scala. Metals is great but some features from idea are not yet there. Metals has a wonderful type checking (except for kind-projector that is not working) which Intellij lacks.

src/main/scala/io/github/pauljamescleary/petstore/infrastructure/endpoint/OrderEndpoints.scala
monadplus: I've already tried to export those type aliases in their own object but it didn't work. I will give a second try because it's a low hanging fruit.

monadplus:
Python tests are  still to do..

pauljamescleary:
@monadplus Could you resolve the merge conflicts?  Apologies for not getting a more thorough review in sooner.

monadplus:
@pauljamescleary sure ! But I am a bit busy next week so I'll try to fix it asap.

pauljamescleary:
@monadplus no worries, thanks for the contrib :)

monadplus:
@paulcleary Branch updated. However, this is going to repeat over and over until python integration tests are added.

src/main/scala/io/github/pauljamescleary/petstore/domain/users/Role.scala
```diff
@@ -0,0 +1,19 @@
+package io.github.pauljamescleary.petstore.domain.users
+
+import cats._
+import cats.implicits._
+import tsec.authorization.{AuthGroup, SimpleAuthEnum}
+
+sealed case class Role(roleRepr: String)
```
zakpatterson: Do you mean final case class here?

src/main/scala/io/github/pauljamescleary/petstore/domain/users/Role.scala
```diff
@@ -0,0 +1,19 @@
...
+
+  val Admin: Role = Role("Admin")
+  val Customer: Role = Role("Customer")
+
+  override val values: AuthGroup[Role] = AuthGroup(Admin, Customer)
+
+  override def getRepr(t: Role): String = t.roleRepr
+
+  implicit val eqRole: Eq[Role] = Eq.fromUniversalEquals[Role]
+}
```
zakpatterson: Add new line here and in more places

src/main/scala/io/github/pauljamescleary/petstore/domain/users/UserService.scala
```diff
@@ -14,7 +14,7 @@ class UserService[F[_]: Monad](userRepo: UserRepositoryAlgebra[F], validation: U
     } yield saved
 
   def getUser(userId: Long): EitherT[F, UserNotFoundError.type, User] =
-    EitherT.fromOptionF(userRepo.get(userId), UserNotFoundError)
+    EitherT.fromOptionF(userRepo.get(userId).value, UserNotFoundError)
```
zakpatterson: `OptionT` transforms to `EitherT` via `toLeft` and `toRight` on `OptionT`.

src/main/scala/io/github/pauljamescleary/petstore/domain/users/UserRepositoryAlgebra.scala
```diff
@@ -1,11 +1,13 @@
...
 
+import cats.data.OptionT
+
 trait UserRepositoryAlgebra[F[_]] {
   def create(user: User): F[User]
 
   def update(user: User): F[Option[User]]
 
-  def get(userId: Long): F[Option[User]]
+  def get(userId: Long): OptionT[F, User]
```
zakpatterson: If we're making this change for this method, we should make the same change for the other two similar ones.

src/main/scala/io/github/pauljamescleary/petstore/domain/users/UserService.scala
zakpatterson: Appreciate you working with what you had here, and I wrote this `EitherT[F, UserNotFoundError.type, User]` thing last year apparently. I'm trying to figure out if this API makes sense now. Why have `EitherT[F, L, R]` where L is fixed to a `case object` type? This `L` provides no extra information compared to `OptionT.none`. Perhaps we should just use this opporunity to simplify this api too. Let me know what you think.

src/main/scala/io/github/pauljamescleary/petstore/infrastructure/endpoint/OrderEndpoints.scala
zakpatterson: Probably a package object for the endpoint package would work

src/main/scala/io/github/pauljamescleary/petstore/infrastructure/endpoint/OrderEndpoints.scala
```diff
@@ -23,40 +30,45 @@ class OrderEndpoints[F[_]: Effect] extends Http4sDsl[F] {
...
   implicit val orderDecoder = jsonOf[F, Order]
 
-  def placeOrderEndpoint(orderService: OrderService[F]): HttpRoutes[F] =
-    HttpRoutes.of[F] {
-      case req @ POST -> Root / "orders" => {
+  def placeOrderEndpoint(orderService: OrderService[F]): AuthEndpoint = {
+      case req @ POST -> Root / "orders" asAuthed user =>
         for {
-          order <- req.as[Order]
+          order <- req.request.as[Order]
```
zakpatterson: line break

src/test/scala/io/github/pauljamescleary/petstore/infrastructure/endpoint/PetEndpointsSpec.scala
```diff
@@ -1,20 +1,19 @@
 package io.github.pauljamescleary.petstore
 package infrastructure.endpoint
 
+import io.github.pauljamescleary.petstore.domain.users._
```
zakpatterson: domain.users._

src/main/scala/io/github/pauljamescleary/petstore/infrastructure/endpoint/OrderEndpoints.scala
zakpatterson: I see, if you factor out the type aliases, you would still have to parameterize just by the `Auth` type, which I think is fine.

zakpatterson:
@monadplus I made some simple suggestions on your PR, this is great work. I haven't really used JWT yet, so I'll have a go at updating the python integration testing to make this pass.

monadplus:
@zakpatterson Thanks for the review. I'll try to reply to the comments and fix them asap 👍

src/main/scala/io/github/pauljamescleary/petstore/domain/users/Role.scala
monadplus: Final suits better here.

src/main/scala/io/github/pauljamescleary/petstore/domain/users/Role.scala
monadplus: I checked if this could be added as a scalafmt rule but seems not to be implemented https://scalameta.org/scalafmt/docs/configuration.html#newlines

src/main/scala/io/github/pauljamescleary/petstore/domain/users/UserService.scala
monadplus: For this use case, having EitherT doesn't add value in the short time. But, I'd rather not change it as if we want to add new errors on the future we will have to change all instances of OptionT to EitherT. Also, replacing EitherT for OptionT won't simplify  the api much..

src/main/scala/io/github/pauljamescleary/petstore/domain/users/UserRepositoryAlgebra.scala
monadplus: I tried to make the lesser possible changes but I think this should be done to share a common codification and having OptionT is better if you ever want to change your monadic stack.

monadplus: BTW, there's no ListT but it can be achieved using Nested.

zakpatterson:
Hi @monadplus 

There is an issue with the manner in which the services are combined, which causes routes to unexpectedly require authentication. I think there are a couple easy fix options for this. 

Here is the issue:

Symptom
---

```
/home/projects/oss/scala-pet-store · (feature/auth±) at  zak-lenovo
⟩ curl -v -X POST http://localhost:8080/users -d @user.json --header "Content-Type: application/json"
Note: Unnecessary use of -X or --request, POST is already inferred.
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 8080 (#0)
> POST /users HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.61.0
> Accept: */*
> Content-Type: application/json
> Content-Length: 116
> 
* upload completely sent off: 116 out of 116 bytes
< HTTP/1.1 401 Unauthorized
< Date: Sun, 05 May 2019 18:03:52 GMT
< Content-Length: 0
< 
* Connection #0 to host localhost left intact
⏎                                                                                                                                    
/home/projects/oss/scala-pet-store · (feature/auth±) at  zak-lenovo
⟩ cat user.json 
{
    "userName": "a",
    "firstName": "b",
    "lastName": "c",
    "email": "d",
    "password": "e",
    "phone": "f"
}
⏎                                                                                                                                    
/home/projects/oss/scala-pet-store · (feature/auth±) at  zak-lenovo
⟩ 
```

Source
---
So as you can see, the signup route right now is throwing 401 unauthorized. Let's see why.

At this line [UserEndpoints class](https://github.com/monadplus/scala-pet-store/blob/feature/auth/src/main/scala/io/github/pauljamescleary/petstore/infrastructure/endpoint/UserEndpoints.scala#L130), you correctly distinguish between unauthenticated routes and authenticated routes.

However, in [the Server class](https://github.com/monadplus/scala-pet-store/blob/feature/auth/src/main/scala/io/github/pauljamescleary/petstore/Server.scala#L40), the `PetEndpoints` are started first, and the `UserEndpoints` are combined later. PetEndpoints require authentication, so every route following the `<+>` combination will also require authentication.

Fix
---
I see this being fixed in one of a couple ways.
* We could divide our services in [the Router declaration](https://github.com/monadplus/scala-pet-store/blob/feature/auth/src/main/scala/io/github/pauljamescleary/petstore/Server.scala#L43), then each endpoint in each endpoints class would have to be updated. In `UserEndpoints`, it would just be `Root` instead of `Root / "users"`, for example.
* At [this point in the Server object](https://github.com/monadplus/scala-pet-store/blob/feature/auth/src/main/scala/io/github/pauljamescleary/petstore/Server.scala#L40), you could just peer into each endpoints, and combine their unauthEndpoints together, then combine their authEndpoints together.

Ok let me know if this makes sense.

monadplus:
I like the second one. What about creating a new type that holds both All and Admins endpoints with a instance of semigroupK that combines them properly ?. This new type will only have an `apply` and `def to: HttpRoutes[F]` methods.

The solution will be very similar to what we already have but the implementation will be different.

```scala
services       =  (PetEndpoints.endpoints[F, HMACSHA256](petService, routeAuth) <+>
OrderEndpoints.endpoints[F, HMACSHA256](orderService, routeAuth) <+> UserEndpoints.endpoints[F, BCrypt, HMACSHA256](userService,BCrypt.syncPasswordHasher[F], routeAuth)).toHttpRoute[F]
```

What do you think @zakpatterson ?

zakpatterson:
If I'm understanding right, what you're suggesting is to have `A <+> B` automatically expand into `A.unauth <+> A.auth <+> B.unauth <+> B.auth`, but you still have the same problem with `B.unauth` being combined on top of `A.auth`.

I am interested in seeing your solution worked out to understand the semigroupK concept better. 

Can you consider the Router option again? it would just be 
```
Router(
   "/pets" -> petEndpoints.endpoints,
   "/users" -> userEndpoints.endpoints,
   "/orders" -> orderEndpoints.endpoints
)
```

monadplus:
@zakpatterson Embarrassingly easy, let's do it this way. Moreover, it's easier to understand.

The idea behind the semigroupK would be combining the auth first and then fallback to the unauth and finally to route  not found. But I don't like it anymore. BTW, all endpoints are authenticated, you must be registered and logged in order to access them, although, some of them are only allowed to admin users.

zakpatterson:
Yeah the fallback from auth -> unauth was explicitly avoided in tsec-http4s. Their rationale was to avoid crawling admin routes as an unauthenticated user. The result is that even this happens:

```
⟩ curl -i http://localhost:8080/oranges -d @user.json
HTTP/1.1 401 Unauthorized
Date: Sun, 05 May 2019 18:45:20 GMT
Content-Length: 0
```
So once an authenticated route is considered, you never get 404.

monadplus:
@zakpatterson The router solution doesn't work. Have a look at the definition and you'll find the answer. It still uses <+> under the hood.

zakpatterson:
Hm you are right indeed. I think perhaps this has worked for me in the past by accident. Thanks for pointing that out! In this case you should be able to get by simply putting the "/users" route as the first argument to router, but it would be nice to include a comment or something since the issue is disguised in this way.

It sure would be nice to find a more explicit treatment of this common problem.

monadplus:
I spent the whole afternoon working on it but still fails. I wrote a test to test the router with all endpoints but some routes are not accessible even when you are authenticated as an Admin. Also, I successfully sign up, log in and list  all pets but I couldn't manage to list the orders. 

I don't have more time during this week so feel free to work on it.

zakpatterson:
Thanks @monadplus for your efforts, will take a look.

zakpatterson:
Hi @monadplus, I updated the python functional testing to handle the bearer tokens I only found one bug in your code causing tests to fail (bd2db0c), as the functional testing was expecting user data to be returned on login. So we should give this one more quick review and call it done!

monadplus:
@zakpatterson Sweet! Let me do quick review during this afternoon. Great work.

zakpatterson:
Spoke a little too soon, I was only running the user_tests, will have to take another look at the python side

src/main/scala/io/github/pauljamescleary/petstore/Server.scala
```diff
@@ -4,16 +4,21 @@ import config._
...
 import cats.effect._
 import cats.implicits._
-import org.http4s.server.{Server => H4Server, Router}
+import org.http4s.server.{Router, Server => H4Server}
 import org.http4s.server.blaze.BlazeServerBuilder
 import org.http4s.implicits._
 import tsec.passwordhashers.jca.BCrypt
 import doobie.util.ExecutionContexts
 import io.circe.config.parser
+import io.github.pauljamescleary.petstore.domain.authentication.Auth
```
zakpatterson: fix import

src/main/scala/io/github/pauljamescleary/petstore/Server.scala
```diff
@@ -4,16 +4,21 @@ import config._
 import domain.users._
 import domain.orders._
 import domain.pets._
-import infrastructure.endpoint.{OrderEndpoints, PetEndpoints, UserEndpoints}
-import infrastructure.repository.doobie.{DoobieOrderRepositoryInterpreter, DoobiePetRepositoryInterpreter, DoobieUserRepositoryInterpreter}
+import io.github.pauljamescleary.petstore.infrastructure.endpoint.PetEndpoints
```
zakpatterson: fix import. should just be `infrastructure...`

src/main/scala/io/github/pauljamescleary/petstore/Server.scala
```diff
@@ -4,16 +4,21 @@ import config._
...
+//import infrastructure.endpoint.{OrderEndpoints, PetEndpoints, UserEndpoints}
```
zakpatterson: remove

src/main/scala/io/github/pauljamescleary/petstore/domain/orders/OrderService.scala
```diff
@@ -9,7 +9,8 @@ import cats.data.EitherT
 class OrderService[F[_]](orderRepo: OrderRepositoryAlgebra[F]) {
   import cats.syntax.all._
 
-  def placeOrder(order: Order): F[Order] = orderRepo.create(order)
+  def placeOrder(order: Order): F[Order] = orderRepo.
+    create(order)
```
zakpatterson: bad line break

src/main/scala/io/github/pauljamescleary/petstore/domain/users/User.scala
```diff
@@ -1,11 +1,22 @@
...
+    id: Option[Long] = None,
+    role: Role
 )
+
+object User {
+  implicit def authRole[F[_]](implicit F: Applicative[F]): AuthorizationInfo[F, Role, User] =
+    new AuthorizationInfo[F, Role, User] {
+      def fetchInfo(u: User): F[Role] = F.pure(u.role)
+    }
+}
```
zakpatterson: new line

src/main/scala/io/github/pauljamescleary/petstore/domain/users/UserValidationInterpreter.scala
```diff
@@ -4,22 +4,25 @@ package users
...
 import cats.data.EitherT
 import cats.implicits._
+import Function._
 
 class UserValidationInterpreter[F[_]: Monad](userRepo: UserRepositoryAlgebra[F]) extends UserValidationAlgebra[F] {
-  def doesNotExist(user: User) = EitherT {
-    userRepo.findByUserName(user.userName).map {
-      case None => Right(())
-      case Some(_) => Left(UserAlreadyExistsError(user))
+  def doesNotExist(user: User): EitherT[F, UserAlreadyExistsError, Unit] =
```
zakpatterson: looks like this doesn't change anything, but this method and others in this file should be simplified, using `.as` and `.toLeft` from `OptionT`.

src/main/scala/io/github/pauljamescleary/petstore/infrastructure/endpoint/OrderEndpoints.scala
```diff
@@ -7,56 +7,62 @@ import io.circe.syntax._
 import org.http4s._
 import org.http4s.circe._
 import org.http4s.dsl.Http4sDsl
-import scala.language.higherKinds
 
+import scala.language.higherKinds
 import domain.OrderNotFoundError
+import io.github.pauljamescleary.petstore.domain.authentication.Auth
```
zakpatterson: fix import statement

src/main/scala/io/github/pauljamescleary/petstore/infrastructure/endpoint/OrderEndpoints.scala
```diff
@@ -7,56 +7,62 @@ import io.circe.syntax._
...
 
-class OrderEndpoints[F[_]: Effect] extends Http4sDsl[F] {
-
-  /* Need Instant Json Encoding */
-  import io.circe.java8.time._
+class OrderEndpoints[F[_]: Effect, Auth: JWTMacAlgo] extends Http4sDsl[F] {
 
-  /* Needed for service composition via |+| */
-  import cats.implicits._
+  import cats.implicits._ // |+|
```
zakpatterson: remove comment

src/main/scala/io/github/pauljamescleary/petstore/infrastructure/endpoint/OrderEndpoints.scala
```diff
@@ -7,56 +7,62 @@ import io.circe.syntax._
...
+  import alias._
 
   /* Needed to decode entities */
-  implicit val orderDecoder = jsonOf[F, Order]
+  implicit val orderDecoder: EntityDecoder[F, Order] = jsonOf[F, Order]
```
zakpatterson: i like adding the type annotation here, and it means you can drop the ascriptions from `jsonOf`.

src/main/scala/io/github/pauljamescleary/petstore/infrastructure/endpoint/alias.scala
```diff
@@ -0,0 +1,11 @@
+package io.github.pauljamescleary.petstore
+package infrastructure.endpoint
+
+import io.github.pauljamescleary.petstore.domain.users.User
+import org.http4s.Response
+import tsec.authentication.{AugmentedJWT, SecuredRequest, TSecAuthService}
+
+package object alias {
```
zakpatterson: why not make this `package object endpoint`, and then you wouldn't have to `import alias._` within the endpoint package.

src/main/scala/io/github/pauljamescleary/petstore/infrastructure/endpoint/alias.scala
```diff
@@ -0,0 +1,11 @@
+package io.github.pauljamescleary.petstore
+package infrastructure.endpoint
+
+import io.github.pauljamescleary.petstore.domain.users.User
```
zakpatterson: should be `import domain.users.User`

src/main/scala/io/github/pauljamescleary/petstore/infrastructure/endpoint/alias.scala
```diff
@@ -0,0 +1,11 @@
+package io.github.pauljamescleary.petstore
+package infrastructure.endpoint
```
zakpatterson: remove `endpoint`, see below, this package object should be defined as endpoint so its contents are available within the endpoint package.

src/test/scala/io/github/pauljamescleary/petstore/Arbitraries.scala
```diff
@@ -1,15 +1,21 @@
...
-import domain.{orders, pets}
-import domain.pets._
-import domain.pets.PetStatus._
-import domain.users._
 import java.time.Instant
+import cats.effect.IO
+import io.github.pauljamescleary.petstore.domain.authentication.SignupRequest
 import org.scalacheck._
 import org.scalacheck.Arbitrary.arbitrary
+import io.github.pauljamescleary.petstore.domain.orders._
```
zakpatterson: fix import statements. should just be `import domain...`

src/test/scala/io/github/pauljamescleary/petstore/infrastructure/endpoint/AllEndpointsSpec.scala
```diff
@@ -0,0 +1,77 @@
+package io.github.pauljamescleary.petstore.infrastructure.endpoint
+
+import cats.effect._
+import io.github.pauljamescleary.petstore.PetStoreArbitraries
+import io.github.pauljamescleary.petstore.domain.authentication.SignupRequest
```
zakpatterson: fix imports

src/test/scala/io/github/pauljamescleary/petstore/infrastructure/endpoint/AllEndpointsSpec.scala
```diff
@@ -0,0 +1,77 @@
+package io.github.pauljamescleary.petstore.infrastructure.endpoint
```
zakpatterson: break the package statement apart to make the import statements below cleaner.

src/test/scala/io/github/pauljamescleary/petstore/infrastructure/endpoint/AuthTest.scala
```diff
@@ -0,0 +1,32 @@
+package io.github.pauljamescleary.petstore.infrastructure.endpoint
```
zakpatterson: break package statement apart to make imports below cleaner.

src/test/scala/io/github/pauljamescleary/petstore/infrastructure/endpoint/util/LoginTest.scala
```diff
@@ -0,0 +1,48 @@
+package io.github.pauljamescleary.petstore.infrastructure.endpoint.util
+
+import cats.data.Kleisli
+import cats.effect.IO
+import io.github.pauljamescleary.petstore.domain.authentication.{LoginRequest, SignupRequest}
```
zakpatterson: fix imports.

src/test/scala/io/github/pauljamescleary/petstore/infrastructure/repository/doobie/AuthQueryTypeCheckSpec.scala
```diff
@@ -0,0 +1,36 @@
+package io.github.pauljamescleary.petstore.infrastructure.repository.doobie
```
zakpatterson: fix imports

zakpatterson:
Great work @monadplus , almost there!

monadplus:
@zakpatterson I'll work on the comments during this afternoon. I let you know when the changes are done.

src/main/scala/io/github/pauljamescleary/petstore/domain/users/UserValidationInterpreter.scala
monadplus: I favour simplicity and easy to understand code over shortness.
And I think the code from below is harder to understand than the one we already have. I am not a big fan of this kind of not primitive operations.
```scala
def doesNotExist(user: User): EitherT[F, UserAlreadyExistsError, Unit] =
    userRepo
      .findByUserName(user.userName)
      .map(UserAlreadyExistsError)
      .toLeft(())
```

monadplus: I will change it for the latter though.

src/main/scala/io/github/pauljamescleary/petstore/infrastructure/endpoint/alias.scala
monadplus: How sure you are about that ? I tried renaming alias for endpoint and you still need to import endpoint._ for the aliases.

src/test/scala/io/github/pauljamescleary/petstore/infrastructure/endpoint/AllEndpointsSpec.scala
monadplus: The whole classpath is required here..

src/test/scala/io/github/pauljamescleary/petstore/infrastructure/endpoint/AuthTest.scala
monadplus: Done

src/test/scala/io/github/pauljamescleary/petstore/infrastructure/endpoint/AllEndpointsSpec.scala
monadplus: After I split the package in two, I no longer need the whole classpath...interesting.

zakpatterson: Yea nested package declarations are neat.

zakpatterson:
@monadplus This is awesome, and almost done. @pauljamescleary I had to rewrite a bunch of the python testing to get the bearer tokens and authenticated/unauthenticated/permissioned/unpermissioned tests to work. Would you mind giving at least the python part of this PR a quick review?  I'll assign you.

monadplus:
@zakpatterson I'll changes requested done. From my part I think we are done. Let me know if you need anything else. I'll be working on other side projects.

functional_test/live_tests/pets_test.py
```diff
@@ -90,18 +105,24 @@ def test_update_pet(pet_store_client):
...
+    admin_context()
+    saved_pet_resp = pet_store_client.create_pet(new_pet)
+    saved_pet = saved_pet_resp.json()
+    customer_context()
+    bad_update = pet_store_client.update_pet(dict(saved_pet, **{"status": "Adopted"}))
+    assert_that(bad_update.status_code, is_(401))
+    admin_context()
+    pet_store_client.delete_pet(saved_pet)
+
+# def test_customer_cannot_delete_pet(pet_store_client, customer_context, admin_context)
```
pauljamescleary: Can we delete this comment?

pauljamescleary:
@monadplus this is incredible.  I have one small comment and then lets get this merged!

functional_test/live_tests/pets_test.py
zakpatterson: just implemented the test.

src/test/scala/io/github/pauljamescleary/petstore/infrastructure/endpoint/AllEndpointsSpec.scala
```diff
@@ -0,0 +1,77 @@
...
+  val ordersEndpoint = OrderEndpoints.endpoints[IO, HMACSHA256](orderService, auth.securedRqHandler)
+
+  val routes = Router(
+    ("/users", usersEndpoint),
+    ("/pets", petEndpoint),
+    ("/orders", ordersEndpoint),
+  ).orNotFound
+
+  // TODO: POST -> /pets access is forbidden
+  ignore("login and post order") {
```
zakpatterson: This is still ignored, I'm trying to finish this up now, and if I can get it in the next few minutes I'll merge.

zakpatterson:
Hooray, great work @monadplus !

PR #166, opened 2019-04-27
Update scala version, remove verify step

zakpatterson:
Verify seems to just re-run the 2.12 build

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,13 @@
 language: scala
 
 scala:
-   - 2.12.6
-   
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,15 +1,13 @@
...
-   - 2.12.6
-   
+   - 2.12.8
+
 python:
   - "2.7"
 
 jobs:
   include:
-    - stage: verify
```
zakpatterson: was there some reason this was needed?

PR #168, opened 2019-05-07
Fixed `findByTag` logic and added unit test

m99coder:
While digging deep into that beautiful project and producing this [step-by-step](https://github.com/m99coder/scala-pet-store-step-by-step) guide, I realized that `findByTag` for the `PetRepositoryInMemoryInterpreter` will not work as it compares a `String` with a `Set[String]`. I tried to come up with a reasonable test for that and a fix for the logic.


src/main/scala/io/github/pauljamescleary/petstore/infrastructure/repository/inmemory/PetRepositoryInMemoryInterpreter.scala
```diff
@@ -44,7 +44,7 @@ class PetRepositoryInMemoryInterpreter[F[_]: Applicative] extends PetRepositoryA
     cache.values.filter(p => statuses.exists(_ == p.status)).toList.pure[F]
 
   def findByTag(tags: NonEmptyList[String]): F[List[Pet]] =
-    cache.values.filter(p => tags.exists(_ == p.tags)).toList.pure[F]
+    cache.values.filter(p => p.tags.exists(pt => tags.exists(_ == pt))).toList.pure[F]
```
pauljamescleary: yea, that definitely looks like a bug.  The nested `exists` is concerning here.

A better solution maybe to have a `tags` index, where each tag in the index points to a `NonEmptyList[String]`?

m99coder: What should be found is the intersection of the tags within a `Pet` (`Set[String]`) and the filter predicate (`NonEmptyList[String]`), right? Actually the method name `findByTag` sounds like initially the parameter was intended to be just one tag (a `String`). Maybe the logic could be simplified in that way?

pauljamescleary: I think this can be slightly simplified by converting the param to a `Set` one time before the filter (as opposed to the `toList` in the filter)...

```scala
    val tagSet = tags.toList.toSet
    cache.values.filter { p => p.tags.intersect(tagSet).nonEmpty }.toList.pure[F]
```

zakpatterson: Avoiding computation of the set intersection would be a tiiiiny bit faster?
```
val tagSet = tags.toNes  // NonEmptyList => NonEmptySet
cache.values.filter(_.tags.exists(tagSet contains_ _)).toList.pure[F]
```

pauljamescleary: oh man, too many underscores, can we at least add parens there :)

src/main/scala/io/github/pauljamescleary/petstore/infrastructure/repository/inmemory/PetRepositoryInMemoryInterpreter.scala
```diff
@@ -43,8 +43,10 @@ class PetRepositoryInMemoryInterpreter[F[_]: Applicative] extends PetRepositoryA
   def findByStatus(statuses: NonEmptyList[PetStatus]): F[List[Pet]] =
     cache.values.filter(p => statuses.exists(_ == p.status)).toList.pure[F]
 
-  def findByTag(tags: NonEmptyList[String]): F[List[Pet]] =
-    cache.values.filter(p => tags.exists(_ == p.tags)).toList.pure[F]
+  def findByTag(tags: NonEmptyList[String]): F[List[Pet]] = {
+    val tagSet = tags.toNes
+    cache.values.filter(_.tags.exists(tagSet contains_ _)).toList.pure[F]
```
pauljamescleary: I am good with this, makes sense.  Can we add some parens in that.  It is really tough to read.

Other than that, I am good, nice job

m99coder: Thank you very much 😃

PR #169, opened 2019-05-15
Upgrade to python3

zakpatterson:
This went smoothly, with just a few suggestions from 2to3

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ scala:
    - 2.12.8
 
 python:
-  - "2.7"
+  - "3.6.8"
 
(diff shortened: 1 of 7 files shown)


pauljamescleary:
Nice @zakpatterson , looks like perhaps there are just some dependencies that need to be updated in the build?

zakpatterson:
> Nice @zakpatterson , looks like perhaps there are just some dependencies that need to be updated in the build?

Ah you showed up just as I was force pushing those updates... we'll see if it works.

zakpatterson:
@pauljamescleary good?

Issue #141, opened 2019-02-05
add Scala 2.13 crossbuild


SethTisue:
I can easily imagine you have dependencies that aren't published for 2.13.0-M5 yet, but adding the cross build and seeing where it gets stuck would be a way to find out what's missing and then we can go nag those maintainers

SethTisue:
hmm, one obvious blocker is going to be http4s

maybe subscribe to https://github.com/http4s/http4s/issues/2025 and once that one gets closed as fixed, swing into action?

pauljamescleary:
@SethTisue can definitely run a quick test to see what issues we will have

pauljamescleary:
@SethTisue  - FYI - https://github.com/pauljamescleary/scala-pet-store/pull/142

PR #175, opened 2019-05-19
Use less specific type classes whenever possible

rmsam:
A tiny little change 🙃

For these two scenarios, a Functor seems to be just enough, so by following "the principle of least power" this can make sense.

I noticed that for some services the context bounds (or implicit evidences) are defined at the class level, but on some of the other ones they are define at the method level. I don't really know if this is something you plan to change (for consistency sake) or whether you'd like to show that it is possible to have these two ways of approching the problem. The reason I mentioned this is that this PR may not make sense (in case you want to standardise on having everything at the class level), so feel free to decline it 😅


zakpatterson:
Regarding the context bounds vs implicit parameters, I am sloppy about this and have no clear established best practice. The other open pull request for 2.13 adds kind projector, which makes the context bounds syntax nicer in some situations.

rmsam:
Hi, you are right! The Validation interpreters can use an Applicative instead of a Monad. I'm going to push these changes and rename the PR to something more generic.

In regards to the best practices, unfortunatelly I can't be very useful here :) I'm still trying to get my head around Cats and Cats Effect (specially around understanding how a business app would use these libraries) and I've been studying this project and am planning on jump to Scala Steward afterwards :)

So, my approach hasn't been very sophisticated. I just try to look at the code for something that may get my attention, replace a type class for a less specific one and if it compiles and the tests run, then I assume it is okay. 

This seems to be easier for type classes defined in Cats, but I don't know if this approach would change the semantics when I follow it for Cats Effect type classes. Let me give an example. Currently, `OrderEndpoints`, `PetEndpoints` and `UserEndpoints` require `Effect`. However, if I replace it with `Sync` (required by Circe's `jsonOf`) the project compiles and all tests pass. Since `Effect` extends `Async` which in turn extends `Sync` it compiles. However, is this a legit change? I don't know :/

src/main/scala/io/github/pauljamescleary/petstore/infrastructure/repository/doobie/DoobieUserRepositoryInterpreter.scala
```diff
@@ -53,7 +53,7 @@ private object UserSQL {
   """.query
 }
 
-class DoobieUserRepositoryInterpreter[F[_]](val xa: Transactor[F])(implicit M: MonadError[F, Throwable])
+class DoobieUserRepositoryInterpreter[F[_]](val xa: Transactor[F])(implicit M: Monad[F])
```
zakpatterson: This is fine. Going back to our discussion about parameters vs context bounds, I think this is a case when context bounds are appropriate.

zakpatterson:
> However, if I replace it with `Sync` (required by Circe's `jsonOf`) the project compiles and all tests pass. Since `Effect` extends `Async` which in turn extends `Sync` it compiles. However, is this a legit change? I don't know :/

I think this is the same as Monad vs Functor. Monad extends Applicative extends Apply extends Functor, so if Functor is all one needs, then that is all one should indicate.

So Sync sounds right given your explanation.

src/main/scala/io/github/pauljamescleary/petstore/infrastructure/repository/doobie/DoobieUserRepositoryInterpreter.scala
zakpatterson: Note neatly using MonadError would have required kind-projector.

rmsam:
@zakpatterson: Completely missed the notifications about your comments :/ Thanaks for them btw :) Meanwhile I've pushed the change from `Effect` to `Sync`.

zakpatterson:
@rmsam, the scala 2.13 crossbuild #141 involved upgrading doobie which brought some changes of constraints on `F`. These merge conflicts look minimal, try to take another look if you can. Thank you!

rmsam:
@zakpatterson: I think that with these changes does PR doesn't make sense anymore :)

zakpatterson:
@rmsam, I'm sorry, I should have merged your branch first before the 2.13 one. I disagree though, this PR is still really good, and helped me find a few instances where we applied the `Bracket[?[_], Throwable]` constraint on F where it wasn't necessary. So I went ahead and pushed some updates into your fork of this, to get us to green again. I'll give it another review and merge tomorrow. Thanks!

rmsam:
Oh okay, that makes sense @zakpatterson. I'm going to have a look at your changes now and see if I can spot anything similar 👍

zakpatterson:
This looks pretty good now, I'm going to go ahead and merge it. Definitely open another PR if you notice anything else. Thanks for your contrib!

Also, do open another PR to add yourself to our AUTHORS.md

PR #243, opened 2019-09-16
Add mergify configuration

zakpatterson:
This is a direct copy from [scala-steward's own configuration](https://github.com/fthomas/scala-steward/blob/ce76c18a074b93697f55c44161830a0b2529c75d/.mergify.yml)

If a PR is sent from scala-steward, and passes CI, then it's automatically merged, and the dependency-update label is applied.

This will require @pauljamescleary installing mergify on his github account (free for open source), and enabling on this repo.

[mergify.io](https://mergify.io)

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,0 +1,10 @@
+pull_request_rules:
+  - name: automatically merge scala-steward's PRs
+    conditions:
+      - author=scala-steward
+      - status-success=Travis CI - Pull Request
+    actions:
+      merge:
+        method: merge
+      label:
+        add: [dependency-update]


.mergify.yml
```diff
@@ -0,0 +1,10 @@
+pull_request_rules:
+  - name: automatically merge scala-steward's PRs
+    conditions:
+      - author=scala-steward
+      - status-success=Travis CI - Pull Request
+    actions:
+      merge:
+        method: merge
```
pauljamescleary: I know we haven't set this up here, but I far prefer squash merge, although I am not sure how much of a difference it makes

zakpatterson: Yeah i prefer squash too but i think steward always commits on recent master and rebases open PRs. So this should be the same i think. Well except maybe if I push a commit into stewards PR that makes it succeed. It’s a good point.

PR #246, opened 2019-09-20
Format with scalafmt

jiminhsieh:
Format all of source code and sbt files with scalafmt.

I didn't modify the rule of current scalafmt.conf. Based on the result, we probably need to discuss which rule we need to apply.


build.sbt
```diff
@@ -1,58 +1,57 @@
-organization        := "io.github.pauljamescleary"
-name                := "scala-pet-store"
-version             := "0.0.1-SNAPSHOT"
-scalaVersion        := "2.12.9"
-crossScalaVersions  := Seq("2.12.9", "2.13.0")
+organization := "io.github.pauljamescleary"
+name := "scala-pet-store"
+version := "0.0.1-SNAPSHOT"
+scalaVersion := "2.12.9"
+crossScalaVersions := Seq("2.12.9", "2.13.0")
```
jiminhsieh: Current config is [`align=none`](https://scalameta.org/scalafmt/docs/configuration.html#align-none). I'm thinking we should consider using [`align=more`](https://scalameta.org/scalafmt/docs/configuration.html#align-most). Or we probably could use [`align.tokens`](https://scalameta.org/scalafmt/docs/configuration.html#aligntokens) to handle `=` and `:=` only.

zakpatterson:
Looking at this I think it's about 90% consistent with our ongoing informal style.

Changes I see are:
1. Removes trailing commas
2. Removes consistent alignment in our big for expression and sbt files.
3. Uses method syntax instead of operator syntax for named methods
4. Double indents `extends`, `with` if they appear on a new line
5. Removes {} enclosing a single expression
6. Newlines if `(\n` follows `=`

My votes:
1. No, I like them because if you add a line to a seq, it only creates a one line diff.
2. Yes. For the same reason as 1. Adding a line to a for expression or similar might require reindenting the entire thing.
3. Yes although I'm lazy about this, I often write `.map(CC apply _)`, but for this project yes let's do this.
4. Yes
5. Yes.
6. No. This one comes up when we have doobie queries of the form:
```
def query: Query0 = (
  sql"""
    select blah from foo ...
  """ ++ queryFragment
).query
```
In this case scalafmt is putting the `(` on a new line. To me at least `= (\n` is a quick signal to the reader that what comes next is a simple expression, not a list of statements.  Perhaps `= \n` does the same thing. Would it be better to break up that query function above into multiple statements? i.e. `val q = sql"""...` followed by `(query ++ fragment).query`?

jiminhsieh:
I will try to add `trailingCommas = always` to scalafmt.conf.

About the 6th point, I don't think Scalafmt could fix about that. Scalafix probably could do the trick.

I will rewrite the history of this PR for adding `trailingCommas = always`. I want to format the master branch version instead of my previous change.

zakpatterson:
Also please append your name to AUTHORS.md.

zakpatterson:
LGTM, thanks so much for the contribution!

pauljamescleary:
Thanks for the contrib!  Sorry I was late, this got to me when I was on vacation.

PR #250, opened 2019-09-24
Use new API instead of the deprecated one

jiminhsieh:
[`Uri.uri` was deprecated](https://github.com/http4s/http4s/blob/v0.21.0-M5/core/src/main/scala/org/http4s/Uri.scala#L736). It suggests the API with string interpolation.


zakpatterson:
Great thank you, I briefly took a look at our travis builds. Based on our settings this deprecation warning should have failed the build. How did you see it?

jiminhsieh:
Intellij let me there were deprecations. I don't know why the compiler didn't detect it. Some of the other rules don't have this issue. I tried to add snippets which caused the rules of `-Ywarn-unused:locals` and `-Ywarn-unused:imports` will catch up. It did compile failed.

PR #272, opened 2019-10-23
Adding an FS2 example for simple order processing

pauljamescleary:
Introduces a naive `Queue` as well as an FS2 stream for handling orders in the background.  The FS2 stream will either save the order (for new orders) or delete the order (for cancels).

The processing follows:
1. On Placing an order, the order is sent to the queue for processing to save it
2. On Cancelling an order, the order is sent to the queue for processing to delete it

The processing is purposefully simple and rather unnecessary, if only to illustrate FS2.

- `OrderProcessor` - this is the main FS2 flow that polls for messages on the queue every few seconds, pipes them through a handler, and then pipes them through to delete the message off of the queue.  We do not explicitly handle errors right now as I planned on implementing this for SQS which does a retry after timeout.  I think this example is good enough, open to other things to show off
- `OrderQueueAlgebra` - this is a queue abstraction generalized to any implementation that follows the SQS approach to receive/delete messages.
- `OrderService` - this was modified to use the queue instead of processing directly
- `OrderQueueInMemoryInterpreter` - a naive implementation of a message queue using purely in-memory constructs.
- `OrderQueueEndpoints` - slight adjustment due to changing the signature of `delete` and renaming to `cancel`

### DIFF
--- a/src/main/scala/io/github/pauljamescleary/petstore/Server.scala
+++ b/src/main/scala/io/github/pauljamescleary/petstore/Server.scala
@@ -5,12 +5,7 @@ import domain.users._
 import domain.orders._
 import domain.pets._
 import infrastructure.endpoint._
-import infrastructure.repository.doobie.{
-  DoobieAuthRepositoryInterpreter,
-  DoobieOrderRepositoryInterpreter,
-  DoobiePetRepositoryInterpreter,
-  DoobieUserRepositoryInterpreter,
-}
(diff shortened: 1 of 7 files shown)


pauljamescleary:
**Note:  I did not fix the tests on this yet as I wanted to get a Prototype up for feedback first.  Will update the tests and such once I get some 👀  **

PR #278, opened 2019-11-19
Add jdk8 at ci

jiminhsieh:
* Run unit tests with OpenJDK 8, but we allow it to fail
* Emphasize that project is used Java 11 as default

Here is the picture at Travis CI:
Screen Shot 2019-11-19 at 12 12 00

I guess it will not affect mergify.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,10 @@
 language: scala
 dist: xenial
+
+jdk:
+  - openjdk8
+  - openjdk11
+
 scala:
    - 2.12.9
    - 2.13.0
@@ -11,6 +16,10 @@ script:
   - sbt scalafmtSbtCheck scalafmtCheckAll
   - sbt ++$TRAVIS_SCALA_VERSION test
   - ./build.sh
+
+matrix:
+  allow_failures:
+    - jdk: openjdk8
 
 cache:
   directories:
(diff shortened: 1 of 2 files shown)


jiminhsieh:
Here is the summary of mergify.
Screen Shot 2019-11-19 at 12 55 22

It won't affect mergify.

zakpatterson:
This makes sense to me @jiminhsieh, thanks for the contrib. I looked around for a while to try to figure out a way to make the "allowed failures" a little more conspicuous, such as adding a label to the PR, or returning a comment to the PR on github. I think version updates will be pushed by steward and then mergify will auto merge them with a green checkmark. 

Anyway, just curious if you know of any tool that would raise a flag on github somehow and still allow merging. I'll merge this as is.

pauljamescleary:
This is great @jiminhsieh , thanks!  I learned some new things about travis in here :)

zakpatterson:
I think we can revert the allowed_failures part now due to new tsec release. paging @jiminhsieh?

jiminhsieh:
I am fine with reverting the `allow_failures`. However, I am thinking about not removing the `allow_failures`. The reason is this project is aiming for Java 11. Adding Java 8 is used for catching up on some issues from upstream.

PR #308, opened 2020-01-28
Add tpolecat scalac options plugin

zakpatterson:
This really makes it easier to manage those scalacoptions per scala version.


zakpatterson:
Passing CI and negative lines, what's not to like?

PR #311, opened 2020-01-29
Compose ConnectionIOs and transact only once

mebubo:


zakpatterson:
Great @mebubo thanks for the contribution. I look at this, and it makes me think the whole DoobieXXRepositoryInterpreter could be simplified...

Instead of
```scala
class DoobiePetRepositoryInterpreter[F[_]: Bracket[?[_], Throwable]](val xa: Transactor[F])
    extends PetRepositoryAlgebra[F] 
```

Do you think it could just be:
```scala
class DoobiePetRepositoryInterpreter extends PetRepositoryAlgebra[ConnectionIO]
```
?

PR #315, opened 2020-01-31
Avoid NonEmptyList.fromListUnsafe

mebubo:


src/main/scala/io/github/pauljamescleary/petstore/infrastructure/endpoint/PetEndpoints.scala
```diff
@@ -86,27 +86,31 @@ class PetEndpoints[F[_]: Sync, Auth: JWTMacAlgo] extends Http4sDsl[F] {
...
-    case GET -> Root / "findByTags" :? TagMatcher(Valid(Nil)) asAuthed _ =>
-      BadRequest("tag parameter not specified")
-
-    case GET -> Root / "findByTags" :? TagMatcher(Valid(tags)) asAuthed _ =>
-      for {
-        retrieved <- petService.findByTag(NonEmptyList.fromListUnsafe(tags))
-        resp <- Ok(retrieved.asJson)
-      } yield resp
+    case GET -> Root / "findByTags" :? TagMatcher(Valid(xs)) asAuthed _ =>
+      NonEmptyList.fromList(xs) match {
```
zakpatterson: I usually would use a fold on option. I see other uses in this repo of pattern matching on `Option`, so I think it's fine. There's also `xs.toNel.fold(BadRequest(_)){ case tags => ... }` or something like that.

zakpatterson:
Great catch thank you. I made a note of a way to shorten this syntactically, but the way you've done it is consistent with how it's done in other places in this repo, so I'm just going to merge as is. Nice contrib!

src/main/scala/io/github/pauljamescleary/petstore/infrastructure/endpoint/PetEndpoints.scala
mebubo: Thanks for the hint. Indeed, `fold` makes it more compact than pattern matching.

Also didn't know about the `.toNel` syntax, nice!

PR #380, opened 2020-06-10
Fix SQLPagination to handle parameters in the original query

ACoalzy:


zakpatterson:
Thanks! Sorry this sat out so long, thanks for the fix!

